PR #11, opened 2015-11-07
Suppress naming consistency checks for MonoBehaviours events

MrJul:
This PR suppresses naming warnings on MonoBehaviour's event handlers.

(eg. AnimatorIK has a warning suggesting to rename it to AnimatorIk with the default rules, thus breaking code if applied)


citizenmatt:
So this suppresses naming checks for known names, right? As in, these names are provided by the Unity API?

MrJul:
That's right. It affects all `MonoBehaviour` event handlers ("messages" in the [documentation](http://docs.unity3d.com/ScriptReference/MonoBehaviour.html)), the same ones that are already not marked as unused by the plugin.

citizenmatt:
Nice one, thanks!

Issue #6, opened 2015-09-14
Do not include Assets in namespace for new classes


justinlueders:
Resharper wants to both modify namespaces to include the top level Assets folder as well as add Assets to the namespace of generated classes.  I don't think most people want this.

alexander-mironov:
+1

citizenmatt:
Could you give an example please? What's the folder structure, and what files are affected?

justinlueders:
When creating a class via resharper, it will place it in a namespace according to the folder structure.  The root of the unity folder structure is always Assets.  So all namespaces start with that.  namespace Assets.Foo.Bar instead of the preferred Foo.Bar

citizenmatt:
Cool. That sounds like we can mark the `Assets` root folder as not being a "namespace provider", and all the other folders will be used as part of the namespace instead. You can try this out by selecting the Assets folder, opening the Properties tool window and changing "Namespace Provider" to `false`. Does that solve the issue?

justinlueders:
rad, sounds like it will, my 30 day trial of 2015 just ran out..need to
find my key for version 8 and ill test it. :)

Shorebound:
I've checked the assets folder to false (all three projects link to the same folder, so when you change one all of them do), but this still doesn't work and I can't find a way around it.

I currently have a default namespace set as TeamName.GameName, but it's suggesting TeamName.GameName.Assets through ReSharper, even though there are no other instances of the Assets folder in the solution explorer. 
![output1](https://cloud.githubusercontent.com/assets/10943141/10839123/f22a1e86-7f23-11e5-9a53-7cf88cebd239.PNG)

citizenmatt:
That's the approach you need to take, but I'm surprised it's not working. I've just tried it with a Unity VS2015 project, and ReSharper is correctly suggesting a namespace that contains the default namespace + folders, but not including "Assets". Do you have a screenshot or something that can help debug it? (A repro project would be great!)

citizenmatt:
I'm taking a deeper look at this - what's the default behaviour for Unity? If I create a new project, I get standard assets that include scripts that have all sorts of different namespaces - `UnityStandardAssets._2D`, `UnityStandardAssets.Cameras`, `UnityStandardAssets.CrossPlatformInput.Inspector`, etc.

[...]

Issue #5, opened 2015-09-14
Automatically set language level to C#5


citizenmatt:
Unity is compiled with mono, which doesn't support C#6, so ReSharper shouldn't use or suggest C#6 features, when running mono projects in VS2015.

The plugin should detect when a mono solution has been opened, and automatically set language level to C#5 for all projects.

citizenmatt:
Unity will also be updating their project generation code to include the `LanguageLevel` msbuild property (or whatever it's actually called). ReSharper pays attention to this, and will act accordingly.

I don't know when this is going to happen, and I've lost the tweet that mentioned this too...

PR #12, opened 2015-11-13
In-memory per-project settings

citizenmatt:
Creates an in-memory per-project settings layer for each unity project. The layer is invisible, and cannot be edited, but can be overridden by the normal `.csproj.dotSettings` and `.csproj.dotSettings.user` files.

The following settings are written to the layer, to act as defaults:

* Disables the `Assets` and `Assets\Scripts` folders as namespace providers. ReSharper will no longer suggest `Assets` or `Scripts` should be part of the namespace (#6)
* Defaults Unity projects to C#5 (#5)

### DIFF
--- a/src/resharper-unity/.editorconfig
+++ b/src/resharper-unity/.editorconfig
@@ -1,0 +1,5 @@
+root=true
+
+[*.cs]
+indent_style=space
+indent_size=4
--- a/src/resharper-unity/ProjectModel/Properties/Flavours/UnityProjectFlavor.cs
+++ b/src/resharper-unity/ProjectModel/Properties/Flavours/UnityProjectFlavor.cs
@@ -1,0 +1,14 @@
+﻿using System;
+using JetBrains.ProjectModel.Properties;
+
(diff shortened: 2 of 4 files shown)


citizenmatt:
@tutushkin: Could you review, please? I'm happy with this, but would like a second opinion, especially on the approach.

tutushkin:
@citizenmatt have you tried the IHaveDefaultSettingsStream approach? I suspect it could be project-level.

citizenmatt:
I did look at that, and `IHaveDefaultSettings`, but all the current implementations are `[ShellComponent]`s. I was just about to suggest making a component that was `[ProjectComponent]`, that took in an `IProject` and decided what to do based on that. The consumer of `IHaveDefaultSettings` is `IViewable`, so adding a new component later would work, I guess (the layer would be at a different priority, though). But I can't find `[ProjectComponent]` - has it been removed?

tutushkin:
Ahh, I heard something about the intention about removing the project component, but I couldn't believe it happen so quickly.
Well, If we can't have the project-level default settings stream, you way of approaching the target seems most reasonable.

citizenmatt:
Cool. Thanks. I'll merge and push to live.

PR #29, opened 2016-03-03
Expansion to consider other Unity types and attributes

rorydriscoll:
This commit expands the objects that the extension knows about to include ScriptableObject, Editor, EditorWindow and AssetPostprocessor. I've added all of the events for these types too.

It also marks functions which are referenced by Unity due to attributes as implicitly used.

Note that I appear to be on a later version of the Resharper SDK. I didn't submit the changes to the config file or the project file which relate to this. This means I wasn't able to compile it on the same setup as the master. I've tested it all locally on my own code, but it would be a good idea to test that it compiles correctly on your setup too.


rorydriscoll:
I've noticed that there's another pull request pending with some similar functionality. If this request gets killed due to the other one, then let me know and I'll at least pull in the change to flag functions with specific unity attributes as used.

citizenmatt:
Very cool PR, thanks a lot for submitting. I'm going to close this in favour of #20, which gives us a more data driven approach.

The usage flagging for the attributes is taken care of by external annotations, implemented in bd67def92fdc0bf9f302e31e5a0efcae1cadaed4.

Apologies for not accepting the PR, and for taking so long to sorting it out! Thanks for the PR!

PR #34, opened 2016-05-20
Adding initial annotations for UnityEngine and UnityEditor not null items

bigbadtrumpet:
I added the items that I know for certainty are not null.  Probably could be more added in the future but this gets the most common items that can never be null.

### DIFF
--- a/src/resharper-unity/annotations/UnityEngine.xml
+++ b/src/resharper-unity/annotations/UnityEngine.xml
@@ -8,4 +8,7 @@
   <member name="P:UnityEngine.GameObject.transform">
     <attribute ctor="M:JetBrains.Annotations.NotNullAttribute.#ctor" />
   </member>
+  <member name="P:UnityEngine.Object.name">
+    <attribute ctor="M:JetBrains.Annotations.NotNullAttribute.#ctor" />
+  </member>
 </assembly>
--- a/src/resharper-unity/annotations/UnityEngine.xml
+++ b/src/resharper-unity/annotations/UnityEngine.xml
@@ -1,0 +1,11 @@
+<assembly name="UnityEngine">
+  <member name="P:UnityEngine.Component.transform">
+    <attribute ctor="M:JetBrains.Annotations.NotNullAttribute.#ctor" />
+  </member>
+  <member name="P:UnityEngine.Component.gameObject">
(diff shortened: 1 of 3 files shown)


citizenmatt:
Looks great, thanks! I'll review properly on Monday - I need to catch up with a number of issues and PRs.

Thanks for submitting!

src/resharper-unity/annotations/UnityEngine.xml
```diff
@@ -0,0 +1,14 @@
+<assembly name="UnityEngine">
+  <member name="P:UnityEngine.Component.transform">
+    <attribute ctor="M:JetBrains.Annotations.NotNullAttribute.#ctor" />
```
citizenmatt: Not sure this is accurate. The XML docs for `Component.transform` state:

> The Transform attached to this GameObject (null if there is none attached).

src/resharper-unity/annotations/UnityEngine.xml
```diff
@@ -0,0 +1,14 @@
...
+  </member>
+  <member name="P:UnityEngine.Component.gameObject">
+    <attribute ctor="M:JetBrains.Annotations.NotNullAttribute.#ctor" />
+  </member>
+  <member name="P:UnityEngine.GameObject.transform">
+    <attribute ctor="M:JetBrains.Annotations.NotNullAttribute.#ctor" />
```
citizenmatt: Same here, XML docs, and [website](http://docs.unity3d.com/ScriptReference/GameObject-transform.html) state `transform` can be `null`.

citizenmatt:
Thanks for the pull request! I need the annotations to address other issues, so I'm merging the PR as it stands, but I'm not sure about the `NotNull` attributes that I've commented on above, if you could take a look, please? If you let me know, I can easily edit the two lines.

Thanks!

bigbadtrumpet:
Yea I have posted a question on Unity's forums to ensure that my assumptions are correct.  However it is my understanding (despite the confusing documentation) that the things are marked are correct.  As soon as I get any information to the contrary I will update it with a new pull request!

citizenmatt:
Awesome, thanks. I think I'll comment them out for now, as it goes against the docs. Would be great to get clarification, though.

PR #63, opened 2016-11-07
Unity.Debug methods annotated as AssertionMethods

joshuaoconnor:
All the UnityEngine.Debug.Assert* methods have updated with external
annotations to be AssertionMethods. This allows users to use these
methods instead of the System.Diagnostic.Debug.Assert* methods.
Resharper will identify these methods as halting when their condition is
false.


citizenmatt:
Very cool! I love getting pull requests :) Thanks!

PR #122, opened 2017-03-12
Switch to new csproj model and VS2017

shalupov:


citizenmatt:
Closing due to different approach - keeping multiple project files instead of merging to single.

PR #178, opened 2017-07-11
Minor fixes to Unity3DRider plugin

citizenmatt:
A couple of useful fixes:

* Adds the standard `// <auto-generated>` header to the Unity3DRider.cs file, so that ReSharper doesn't analyse the file. The version check has been updated to still match.
* Removes a `// TODO` comment, which would appear in the user's TODO explorer
* Removes the "Open C# Project in Rider" menu for Unity 2017.1 and above. This was added because Unity would launch both Rider and MonoDevelop with the standard "Open C# Project" menu item. It would do this because it didn't know Rider could handle .sln files properly. Unity 2017.1 recognises Rider, and no longer launches MonoDevelop, so this menu item is no longer necessary. The projects are still processed by the plugin with the standard menu item.


citizenmatt:
I did do that originally :) I made it `#if UNITY_4 || UNITY_5` to match the check in [`InitialiseEditorInstanceJson`](https://github.com/JetBrains/resharper-unity/blob/0419918f2c6b3db48e0b9b5f2905f1bd78806cb4/resharper/src/resharper-unity/Unity3dRider/Assets/Plugins/Editor/JetBrains/RiderPlugin.cs#L201-L201), which is also no longer needed for Unity 2017.1. I'll change them both.

PR #199, opened 2017-08-09
Allow to set desired TargetFrameworkVersion in Unity

van800:
Consider VS2017 MSBuild + dotnet framework targetpack 4.7 will not load project with TargetFrameworkVersion = 4.5, so we need to allow customization.
Unfortunately proper detection of installed targets packs is complicated.

### DIFF
--- a/resharper/src/resharper-unity/Unity3dRider/Assets/Plugins/Editor/JetBrains/RiderPlugin.cs
+++ b/resharper/src/resharper-unity/Unity3dRider/Assets/Plugins/Editor/JetBrains/RiderPlugin.cs
@@ -104,17 +104,13 @@ namespace Plugins.Editor.JetBrains
       get { return EditorPrefs.GetString("Rider_TargetFrameworkVersion", EditorPrefs.GetBool("Rider_TargetFrameworkVersion45", true)?"4.5":"3.5"); }
       set
       {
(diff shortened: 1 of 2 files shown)


van800:
found a bug, do not merge

resharper/src/resharper-unity/Unity3dRider/Assets/Plugins/Editor/JetBrains/RiderPlugin.cs
```diff
@@ -467,16 +480,17 @@ static void RiderPreferencesItem()
...
     - Rider will fail to resolve System.Linq on Mac/Linux
     - Rider will fail to resolve Firebase Analytics.
  - With 4.5 Rider will show ambiguous references in UniRx.
 All those problems will go away after Unity upgrades to mono4.";
-      TargetFrameworkVersion45 =
-        EditorGUILayout.Toggle(
-          new GUIContent("TargetFrameworkVersion 4.5",
-            help), TargetFrameworkVersion45);
+           
+      TargetFrameworkVersion =
```
van800: default here is TargetFrameworkVersion, it has default value in the getter
EditorPrefs.GetBool("Rider_TargetFrameworkVersion45", true)?"4.5":"3.5")

PR #202, opened 2017-08-10
Allow to see found Rider executables and make one  of them a Default External Editor

van800:
On MAC there is a bug in Unity - Recently used external app disappears from the list of possibilities, when you select other Editor.
So on Rider settings Tab we provide am alternative UI to help find and set Rider as default editor.

### DIFF
--- a/resharper/src/resharper-unity/Unity3dRider/Assets/Plugins/Editor/JetBrains/RiderPlugin.cs
+++ b/resharper/src/resharper-unity/Unity3dRider/Assets/Plugins/Editor/JetBrains/RiderPlugin.cs
@@ -550,8 +550,6 @@ All those problems will go away after Unity upgrades to mono4.";
       EditorGUILayout.EndVertical();
     }
 
-    public static bool RiderIsDefaultEditor { get; set; }
-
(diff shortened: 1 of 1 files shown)


resharper/src/resharper-unity/Unity3dRider/Assets/Plugins/Editor/JetBrains/RiderPlugin.cs
```diff
@@ -43,60 +43,47 @@ private static string GetDefaultApp()
...
-                /*Log(LoggingLevel.Verbose, "Update {0} to {1}", alreadySetPath, newPathMac);
-                SetExternalScriptEditor(newPathMac.FullName);*/
-                return newPathMac.FullName;
-              }
-            }           
-            break;
-        }
+    private static string[] GetAllRiderPaths()
+    {
+      switch (SystemInfoRiderPlugin.operatingSystemFamily)
```
IvanPashchenko: nothing for Linux for now?

van800: not needed I think

PR #228, opened 2017-09-11
Shaderlab parser improvements

citizenmatt:
A whole bunch of improvements.

* Unquoted strings in property attributes, including nested parentheses
* Boolean values correctly expect `on`|`off` and `true`|`false` and used in more places
* More places accept variable references
* Parse `ZClip`
* Improve legacy texture binding blocks
* Cg blocks now treated as pre-processor tokens, so can be parsed anywhere (they normally get converted into a `GpuProgramId` command, or multiple passes for surface shaders)
* Handle nested multi-line comments


resharper/src/resharper-unity/Host/Features/Foldings/ShaderLab/ShaderLabCodeFoldingProcessor.cs
```diff
@@ -30,28 +30,20 @@ public override void VisitNode(ITreeNode node, IHighlightingConsumer consumer)
                 ShaderLabTokenType.MULTI_LINE_COMMENT, ShaderLabTokenType.NEW_LINE);
         }
 
+#pragma warning disable 672
+        // Obsolete warning tells us to also implement VisitTexturePropertyValueNode
```
IvanPashchenko: Why are we ignoring the warning? Is `VisitBlockValueNode` not really obsolete?

resharper/src/resharper-unity/Psi/ShaderLab/Parsing/ShaderLab.lex
```diff
@@ -100,6 +97,8 @@ PP_DIGITS={DECIMAL_DIGIT}(({WHITESPACE})*{DECIMAL_DIGIT})*
 <YYINITIAL,PARENS>          ","                 { return ShaderLabTokenType.COMMA; }
 <YYINITIAL>     "."                             { return ShaderLabTokenType.DOT; }
 <YYINITIAL>     "+"                             { return ShaderLabTokenType.PLUS; }
+<YYINITIAL>     "-"                             { return ShaderLabTokenType.MINUS; }
+<YYINITIAL>     "+-"                            { return ShaderLabTokenType.PLUS_MINUS; }
```
IvanPashchenko: When can this be used (just curious)? :)

resharper/src/resharper-unity/Psi/ShaderLab/Parsing/ShaderLab.psi
```diff
@@ -888,52 +1022,75 @@ scalarPropertyValue
...
+  TEXTURE_2D_KEYWORD<SHADER_LAB_KEYWORD, Keyword>
+  | TEXTURE_3D_KEYWORD<SHADER_LAB_KEYWORD, Keyword>
+  | CUBE_KEYWORD<SHADER_LAB_KEYWORD, Keyword>
+  | TEXTURE_2D_ARRAY_KEYWORD<SHADER_LAB_KEYWORD, Keyword>
+  | CUBE_ARRAY_KEYWORD<SHADER_LAB_KEYWORD, Keyword>
+  | RECT_KEYWORD<SHADER_LAB_KEYWORD, Keyword>
+  | ANY_KEYWORD<SHADER_LAB_KEYWORD, Keyword>
+;
+
+// TODO: Shader compiler error - obsolete
```
IvanPashchenko: Technically it should be a warning, not an error (as it is in Unity). Hopefully, the versions of Unity where they are still working are really old.

resharper/src/resharper-unity/Psi/ShaderLab/Parsing/ShaderLab.psi
```diff
@@ -1225,16 +1345,27 @@ cullCommand
...
-  | referencedProperty<SHADER_LAB_REFERENCE, ReferencedProperty>
+  falseLiteral{FALSE_LITERAL, FalseLiteral}
+  | cullOrientation<ORIENTATION, Orientation>
+;
+
+cullOrientation
+:
+  FRONT_KEYWORD<SHADER_LAB_KEYWORD, Keyword>
+  | BACK_KEYWORD<SHADER_LAB_KEYWORD, Keyword>
+  | variableReference<SHADER_LAB_REFERENCE, VariableReference>  // Reference to float, but which value is front and which back?
```
IvanPashchenko: Seems to be Off - 0, Front - 1, Back - 2

image

IvanPashchenko:
LGTM, merge at your own discretion

resharper/src/resharper-unity/Psi/ShaderLab/Parsing/ShaderLab.lex
citizenmatt: It's a [legacy operator](https://github.com/JetBrains/resharper-unity/blob/shaderlab-parser-improvements/resharper/src/resharper-unity/Psi/ShaderLab/Parsing/ShaderLab.psi#L1855), used as part of the `Combine` calculation, same as `dot3` and `dot3rgba`. It needs to be flagged as an error with an inspection.

resharper/src/resharper-unity/Psi/ShaderLab/Parsing/ShaderLab.psi
citizenmatt: I thought Unity did treat this as an error. But anyway, good point. It'll be worth investigating when these constructs became errors/warnings and only highlight them if we're targeting that version or above. We could perhaps make the severity of our "compiler" errors configurable, unlike say C#.

resharper/src/resharper-unity/Host/Features/Foldings/ShaderLab/ShaderLabCodeFoldingProcessor.cs
citizenmatt: It's what happens when an element is used in multiple interfaces - the `Visit{Whatever}Node` gets the `Obsolete` attribute to warn that the overload won't get called for certain element types (and it lists which types). It's not obsolete, but it's a way of getting a warning into the compiler. The overload is still called for elements that only implement the `IBlockValue` interface.

I don't really like it (I'd prefer to compile with warnings as errors on, ideally), but I think I want to keep the multiple interfaces, at least for now. It's useful to be able to treat `TexturePropertyValue` as both a property value (used in `Properties`) and a block value (has `{` and `}`, useful for folding. It might be that as we use the AST for more things, it doesn't make as much sense, and then we can perhaps reconsider. Perhaps `IBlockValue` should move out of the parser, and be added by hand in partial classes.

And I checked the C# parser, and it's got loads of them, too :)

Issue #232, opened 2017-09-14
Allow using Unity script templates in Rider


van800:
We need some understanding if ppl use them outside UnityEditor.
Maybe we can provide them/similar directly in Rider?
Also people may need a helper to update Unity templates or just help to find where are they located.

van800:
http://www.tristancartledge.com/custom-unity-script-templates/

jcx:
in unity, there are just 2 default script template and it's for the basic MonoBehaviour and ScriptableObject

there is also free asset made by unity but it's outdated so it shows error and you need to install it after you create projects so it's pretty annoying and I stopped to use it a long time ago

I also made class creator but similar problem - need to install it to every new project 

also you can make your own templates and put them together with the default but the problem is, it's in unity installation directory for example "c:\Program Files\Unity20171p1\Editor\Data\Resources\ScriptTemplates\.." and that's bad ... if you do need update unity often (and it's actually everybody who's not finishing huge project ... then you just resign on that

anyway as dev you mostly need the MonoBehaviour and the ScriptableObject and we can type so why do I bother, right? :D 

BUT if you need Editor or EditorWindow you have to do it from the scratch and if you don't make tools on daily basis you need to go to reference and just copy it

what unity needs is possibility to create scripts from template stored elsewhere then in the installation folder and actually the best way to create scripts is outside of unity because unity recompiles the project after you change something in it so the way of creating scripts in IDE is the best, because you create script, name and edit it and then save and switch to unity to recompile

unityVS can create scripts but also just MonoBehaviour and ScriptableObject (even if they claim it's editor script, it's SO)

what I expect from the template is:
1. change the class name
2. selecting methods I want to create in it
3. some basic attributes (like Editor connection to its component or ScriptableObject and EditorWindow connected to its menu path
5. it would be really nice to be able to create more class files at once (that's what my tool does)

van800:
@jcx Thanks much! 
As for multiple classes at once feature, it is hard to think about such feature without more details. Maybe to avoid much writing you don't mind sharing a screencast?

jcx:
sure we can do it, I need to find my tool first :D

jcx:
I thought about it and I guess my idea is pretty complicated for the usual user. The way I am thinking we could get to UML and it's completely different topic.

I guess the usual templates would work just fine and all I need is more of them than just the 2 basic runtime classes.

citizenmatt:
[...]

PR #237, opened 2017-09-22
WIP: File templates

citizenmatt:
# Work in progress - do not merge!

Add file templates.

- [x] File template scope for ReSharper. Templates created in this scope are only valid for Unity projects and show in the "Unity" scope in the ReSharper UI (no UI to edit templates in Rider)
- [x] QuickList support, to show file templates only in Unity projects
- [x] Add [template compiler](https://github.com/citizenmatt/resharper-template-compiler) as a dev dependency to make it easier to create/maintain scripts. Currently needs updating to include msbuild targets and wrap in a nuget. Want to avoid committing binaries, if possible.
- [x] Default templates to match existing Unity script templates
- [x] Potentially add additional scopes, so e.g. can only use in the Assets folder, or in Editor projects, or Plugins folder?

Addresses #232, although there isn't yet a UI to edit templates in Rider. If you edit them in ReSharper and save to shared team (or a custom shared layer), the changes will get picked up.


citizenmatt:
Default templates will be based on existing Unity templates.

C# files:
* New MonoBehaviour
* New edit mode test
* New play mode test
* New playable behaviour
* New playable asset

Unity also defines a new state machine behaviour and new sub state machine behaviour scripts, but they don't display in the UI. Both files are just classes that derive from `StateMachineBehaviour` and have the magic methods in the file, but commented out. Not sure if we want to bother with these.

Shaders:
* New standard surfaces shader
* New unlit shader
* New image effect shader
* New compute shader

I don't think ReSharper (and therefore Rider) allows grouping, or separators in the Quick List templates, although it looks like IntelliJ does. That would be nice.

citizenmatt:
OK, I think this is pretty much good to go. But I'm feeling the need to give a little context to some of the implementation and decisions.

This PR adds file templates for ReSharper and Rider. The experience is currently better in R#, because we can edit the templates and the quick list contents. Rider can only consume the defaults.

The file templates are stored as markdown and ["compiled"](https://github.com/citizenmatt/resharper-template-compiler) into a .dotsettings file that's included as a resource and loaded as a set of defaults. You can [browse them here](https://github.com/JetBrains/resharper-unity/tree/feature/file-templates/resharper/src/resharper-unity/Templates#readme). They're all in a "quick list", which means they all appear in the Insert menu, but only when the appropriate scopes are met, so we don't see them for normal C# projects. The only real change I've made to the standard Unity scripts is to remove the empty magic methods in the new [MonoBehaviour template](https://github.com/JetBrains/resharper-unity/blob/feature/file-templates/resharper/src/resharper-unity/Templates/MonoBehaviour.md) - Rider would immediately flag them as redundant.

You can [find the scopes here](https://github.com/JetBrains/resharper-unity/tree/feature/file-templates/resharper/src/resharper-unity/Feature/Services/LiveTemplates/Scope). Scopes are hierarchical, and inherit from earlier states. In other words, as you go up the tree of scopes, they become wider, and narrower when you go down the tree.

```
Unity projects - any folder in any unity project, including root
+ Assets folder - anywhere under Assets in a Unity project, even in subfolders
  + Editor folder - anywhere under an Editor folder (but not first pass)
  + Runtime folder - anywhere that isn't under an Editor folder or first pass
  + Firstpass folder - anywhere that is under Standard Assets, Pro Standard Assets or Plugins
    + Firstpass Editor folder - anywhere under Editor in a first pass folder
    + Firstpass runtime folder - anywhere in a first pass folder that isn't an Editor folder
```

Most of the templates are put in the "Assets folder" scope, so they appear everywhere under `Assets` in a Unity project. But [_Edit Mode Test_](https://github.com/JetBrains/resharper-unity/blob/feature/file-templates/resharper/src/resharper-unity/Templates/EditModeTest.md) is only available in the _Editor folder_ and _firstpass Editor folder_ scopes (a template can be in multiple scopes), so it only appears under an `Editor` folder. The other scopes are currently unused, but could easily prove useful.

I think that basing the scopes on the folder is the best approach, although it's not perfect. The alternative is to try and base them on project type, but I think that's too tricky. Firstly, we'd have to use project/assembly name to identify the project and secondly, the files need to be created in certain folders anyway, so you'd be more likely to get it wrong if it was just based on project. It's still possible to get it wrong - if there's a .txt file in an `Editor` folder, the folder gets included in a runtime project, so we could offer the template there. We could always try adding a project type check to the existing scopes if it's not working well enough, or inspections on the code or something.

When we release this, we could do with asking for feedback on the default templates, ideas for new templates and feedback on the scopes.

resharper/src/resharper-unity/Feature/Services/LiveTemplates/UnityProjectScopeCategoryUIProvider.cs
```diff
@@ -0,0 +1,44 @@
...
+using JetBrains.ReSharper.Feature.Services.LiveTemplates.Scope;
+using JetBrains.ReSharper.Feature.Services.LiveTemplates.Templates;
+using JetBrains.ReSharper.Plugins.Unity.Feature.Services.LiveTemplates.Scope;
+using JetBrains.ReSharper.Plugins.Unity.Resources;
+using JetBrains.ReSharper.Psi.CSharp.Resources;
+
+namespace JetBrains.ReSharper.Plugins.Unity.Feature.Services.LiveTemplates
+{
+    // Defines a category for the UI, and the scope points that it includes
+    [ScopeCategoryUIProvider(Priority = -200, ScopeFilter = ScopeFilter.Project)]
```
IvanPashchenko: What does `-200` mean here? Do we have some constant in SDK to use instead of that? :)

resharper/src/resharper-unity/Feature/Services/LiveTemplates/UnityFileTemplatesSupport.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Collections.Generic;
+using JetBrains.ReSharper.Feature.Services.LiveTemplates.Scope;
+using JetBrains.ReSharper.Feature.Services.LiveTemplates.Support;
+using JetBrains.ReSharper.Plugins.Unity.Feature.Services.LiveTemplates.Scope;
+
+namespace JetBrains.ReSharper.Plugins.Unity.Feature.Services.LiveTemplates
+{
+    // Used when creating templates from legacy XML.
+    // Not really sure if we need it, to be honest
```
IvanPashchenko: What happens if we remove this?

resharper/src/resharper-unity/Feature/Services/LiveTemplates/UnityProjectScopeProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            var project = context.GetProject();
+            if (project == null)
+                yield break;
+
+            if (!project.IsUnityProject())
+                yield break;
+
+            // We could check for C# here, like InRazorCSharpProject, but we only really support C# Unity projects
+            // Are there any other types?
```
IvanPashchenko: Well, there could be pure UnityScript projects, but they are not supported by us right now and UnityScript is going to be turned off  soon anyway AFAIK.

IvanPashchenko: Also, you can have F# project using the plugin https://forum.unity.com/threads/f-kit.411420/

resharper/src/resharper-unity/Feature/Services/LiveTemplates/UnityProjectScopeCategoryUIProvider.cs
citizenmatt: Sadly, no. It just needs to be less than existing categories, so it appears last in the list in the ReSharper Template Editor. Rider seems to group by quick list, possibly alphabetically, but it ignores this. I'll add a comment.

resharper/src/resharper-unity/Feature/Services/LiveTemplates/UnityFileTemplatesSupport.cs
citizenmatt: Too scared to find out :)

I'll see.

resharper/src/resharper-unity/Feature/Services/LiveTemplates/UnityProjectScopeProvider.cs
citizenmatt: Perhaps I shouldn't assume C# here, then

van800:
I can't find Editor test script anywhere.
image
It would also be good to provide editor test script in Editor assebly and Play Test script in the non-Editor one.

citizenmatt:
The editor test script template is only available in the "Editor folder" scope (see [comment above](https://github.com/JetBrains/resharper-unity/pull/237#issuecomment-334118243) for all the available scopes), so it will only be visible in the menu when invoked on the `Editor` folder or below. This means it should only be available in editor assembly projects (as code in `Editor` folders is only included in editor assembly projects).

The Play Mode Test template is actually in the "Assets folder" scope, which means it's available in both runtime and editor projects. We could change this to "Runtime folder" scope, so it would only be available in folders that weren't first pass, and weren't `Editor`. But this means that it won't be available on `Editor`, which might be useful - what do you think?

van800:
@citizenmatt makes sense, thanks for clarification.

PR #247, opened 2017-10-10
Add attribute target usage annotations

citizenmatt:
Some Unity attributes don't set `AttributeTargets`, put they can only realistically be applied to certain targets. This adds the targets in as external annotations - ReSharper treats this as though they were "real" and shows a warning if the attributes are applied to the wrong target type.

### DIFF
--- a/resharper/src/resharper-unity/annotations/UnityEditor.xml
+++ b/resharper/src/resharper-unity/annotations/UnityEditor.xml
@@ -11,6 +11,11 @@
     <attribute ctor="M:JetBrains.Annotations.BaseTypeRequiredAttribute.#ctor(System.Type)">
       <argument>UnityEditor.Editor</argument>
     </attribute>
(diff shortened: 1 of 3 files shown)


citizenmatt:
Let me know when it's good to merge, and where to merge to - this isn't required for Rider 2017.2

mmechtley:
Currently (2.1.3.4269) the severity of this inspection is not exposed in the UI and defaults to Error. We'd like the ability to downgrade this since it's not actually something that will break builds, just an improper use case.

citizenmatt:
Yep. It's confusing that it's marked as an error when it compiles, although we do sometimes mark things that result in a runtime failure as an error, and this does fit into that category (e.g. a property marked with `[SerializeField]` won't serialise) and it should be fixed.

But we're going to change it to display as redundant code (greyed out), as they can technically be applied, but they won't do anything. See #325.

PR #248, opened 2017-10-11
Method signature inspections

citizenmatt:
Add inspections for methods that are marked with the following attributes:

* `OnOpenAsset`
* `DidReloadScripts`
* `PostProcessScene`
* `PostProcessBuild`

Also improves the method signature inspections to check for incorrect type parameters.


resharper/src/resharper-unity/MethodSignature.cs
```diff
@@ -65,5 +64,42 @@ public static MethodSignature AsMethodSignature(this UnityEventFunction eventFun
...
+        public static bool HasMatchingReturnType(this MethodSignature methodSignature,
+            IMethodDeclaration methodDeclaration)
+        {
+            return Equals(methodSignature.ReturnType, methodDeclaration.Type);
+        }
+
+        public static bool HasMatchingTypeParameters(this MethodSignature methodSignature,
+            IMethodDeclaration methodDeclaration)
+        {
+            // We don't have any generic methods. So it's an error if anyone has any
```
IvanPashchenko: I think it would be better to reflect this in the name of the method.

PR #283, opened 2017-11-18
[WIP] Support for assembly definition files

citizenmatt:
**WORK IN PROGRESS. DO NOT MERGE**

Add support for [assembly definition files](https://docs.unity3d.com/2017.3/Documentation/Manual/ScriptCompilationAssemblyDefinitionFiles.html).

- [x] Add JSON schema validation for .asmdef files
- [x] Add File Template to create a new "Assembly Definition File"
- [x] File Template scope for Unity 2017.3+
- [x] Add duplicate item schema validation for .asmdef files. ReSharper should do this for JSON, but it's not implemented yet. See [RSRP-467118](https://youtrack.jetbrains.com/issue/RSRP-467118). Warns of duplication of platform and reference
- [x] Validation of correct `references` tag in JSON. Requires a cache of existing definition files (which might not be included in the project)
- [x] Code completion of `references` elements
- [x] Go to declaration/find usages on `references` elements
- [x] Highlight usages
- [x] Inspection for self-reference + quick fix
- [x] Inspection for name matching file name + quick fix
- [ ] Validation of circular references
- [x] Rename reference
- [x] Rename file when renaming reference
- [x] Reimplement file type integration so that it only applies to asmdef files, not all JSON files
- [x] Ensure .asmdef is added to list of files to be added to .csproj. This is true for new projects, but not for existing projects. It looks like there is code to always include it, but it doesn't work.


IvanPashchenko:
LGTM

note: I've restarted Travis builds, they were failed due to issues on their side

Issue #265, opened 2017-11-06
Rider no longer works with external solution.


ArtOfSettling:
Maybe my situation is quite unique, and I've found a solution, but it would be great if your extension took care of this automatically.

1. I build my code not through the Unity Provided solution, instead I've got a manually prepared solution.
2. When attaching to Unity, Rider fails to attach.
3. Rider presents the following error when attempting to attach : 'Cannot read Library/EditorInstance.json - file is missing'

My directory setup is something like the following ....

/unity/AllUnityCrap
/solution/MyCustomBuiltSolution

citizenmatt:
The Rider plugin needs to know the process ID of the running instance of the Unity editor so it can attach and debug. The plugin that we install into the Unity editor (the `Unity3dRider.cs` file) will automatically create the `EditorInstance.json` (later versions of Unity create the file themselves). If you're not working in a Unity controlled project, no-one's going to be creating that file, so we can't find a process ID to attach to.

However, if we can't automatically get the PID, we should display a dialog listing all currently running Unity editors, and allow you to choose. It will display a message like you describe, but it shouldn't stop you connecting. Is this not happening? In fact, without checking, if there's only one instance of the Unity editor running, I thought it would automatically attach anyway.

ArtOfSettling:
It used to work this way, it doesn't anymore unfortunately, on top of that, I can manually migrate my EditorInstance.json file into the appropriate location (not really a solution, I just did it for testing), thereby allowing rider to know which process to connect to, and it connects fine, however breakpoints report : 'The breakpoint could not yet be bound to a valid location'.

Unfortunately I updated Unity, Rider and the Unity plugin at the same time, so I'm not sure which is to blame, I could do a little bit further investigation on my end, if it could help, or I could provide you any logs, projects, etc that exhibit this issue.

Rider Version : 2017.2 : 172.4144.1459
Unity Version : 2017.1.1f1
Unity Plugin Version (less relevant I guess) : 2.1.2.1402
Mono : 5.4.0

van800:
In Rider Run-Attach to local processes - select Unity doesn't require EditorInstance.json.
You may try to change mono in Rider settings to the one shipped with Unity. Let us know if it helps with debugging.

ArtOfSettling:
Little bit of an update on this, it seems to be an issue with mono > 5.0.0 since mono > 5.0.0 generates pdb files instead of mdb files by default.

Rider complains that : 'Old versions of mono don't support pdb files'

[...]

PR #290, opened 2017-11-24
Integration of pdb2mdb for debug of dll-s

van800:
Note that it works good on windows, but it is impossible for *nix. 
Msbuild on *nix produces Portable Pdb, which can't be processed by pdb2mdb.
However it is good to merge, I think.


van800:
related to https://github.com/JetBrains/resharper-unity/issues/265

citizenmatt:
Some context: When building external assemblies on Windows, the compiler generates debug information as `.pdb` files, but Unity/Mono requires a different file format (`.mdb`) and debugging won't work for that assembly.

This change will run `pdb2mdb` on any `.pdb` files copied into `Assets`, converting the `.pdb` file into a `.mdb` file that Unity's Mono can understand. It only works on Windows, as it only supports "Windows PDBs".

Newer versions of the Mono toolset on Mac/Linux use Microsoft's CSC compiler, which generates `.pdb` files, but these are "Portable PDBs" and are not compatible with `pdb2mdb`. You can use Unity's Mono toolset in Rider, by changing the _Mono executable path_ and "Use MSBuild version" drop downs to select the detected Unity values. This limits external assemblies to C# 6.

**_We need help testing this!_**

Please let us know if there are problems with this workflow.

PR #304, opened 2017-12-07
[WIP] Open external documentation for Unity symbols

### LINKED ISSUE #98: Feature: Action to Launch Unity Docs for selection
This is a lame request, but MonoDevelop and VSCode+UnityPlugin both have a handy shortcut (CMD+' on MacOS) to launch external documentation for your selection. 

I replicated the desired functionality in Rider using the following:

1. Add new "External Tool" that runs `open` in the console and passes it: https://docs.unity3d.com/ScriptReference/30_search.html?q=$SelectedText$ 
2. Record new macro that runs the action `EditorSelectWord` (expands selection from cursor to entire word) and then the action to run my new "External Tool" 

This works fine for me, but new Rider / ReSharper + MSVS users moving from MonoDevelop or VSCode will try that basic shortcut, then look to bind it, and find that the "External Documentation" shortcut doesn't do anything and that QuickDoc, while helpful, doesn't replace that function.

(I couldn't find a way to configure External Documentation in Rider, maybe this is a Rider EAP specific issue?)

citizenmatt:
An alternative approach, instead of waiting for an API in the protocol and in ReSharper.

Sets up a context consumer, similar to ReSharper's context highlighters that gets called when the context changes, e.g. the cursor is on a new symbol. If that symbol is a Unity symbol, pushes the fully qualified name into the `CustomData` protocol element. When the user calls the _External Documentation_ action, or clicks the icon in the QuickDoc popup, a new `DocumentationProvider` on the front end will check `CustomData` for the symbol name. If present, it generates a URL for the docs, either for locally installed docs, or Unity's search page.

Fixes #98 

Still to do:

- [x] Works and tested on Mac
- [x] Test on Windows
- [x] Verify path for local Linux docs

### DIFF
--- a/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/quickDoc/UnityDocumentationProvider.kt
+++ b/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/quickDoc/UnityDocumentationProvider.kt
@@ -4,12 +4,15 @@ import com.intellij.lang.documentation.DocumentationProvider
 import com.intellij.openapi.util.SystemInfo
 import com.intellij.openapi.vfs.LocalFileSystem
 import com.intellij.openapi.vfs.VfsUtil
+import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiManager
 import com.intellij.util.PathUtil
 import com.jetbrains.rider.projectView.solution
 
 class UnityDocumentationProvider : DocumentationProvider {
+
+    private val documentationRoot by lazy(this::findDefaultDocumentationRoot)
 
     override fun getUrlFor(p0: PsiElement?, p1: PsiElement?): MutableList<String>? {
         val data = p0?.project?.solution?.customData?.data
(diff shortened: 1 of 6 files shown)


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/quickDoc/UnityDocumentationProvider.kt
```diff
@@ -0,0 +1,64 @@
...
+            SystemInfo.isWindows -> {
+                // Unity is installed to `C:\Program Files\Unity`, on both 32 bit and 64 bit
+                // %PROGRAMFILES% differs if we're a 32 or 64 bit process
+                // %PROGRAMFILES(X86)% only exists if we're 64 bit, and is always in the form `C:\Program Files (x86)`
+                // %PROGRAMFILESW6432% is always in the form `C:\Program Files`
+                val envvar = System.getenv("ProgramW6432") ?: System.getenv("ProgramFiles")
+                PathUtil.toSystemDependentName(envvar).plus("/Unity/Editor/Data/Documentation/en")
+            }
+            SystemInfo.isMac -> "/Applications/Unity/Documentation/en"
+            SystemInfo.isUnix -> "/opt/Unity/Editor/Data/Documentation/en"
```
citizenmatt: @van800: Is this the correct location for documentation on Linux?

van800: Looks like Documentation is not included in Unity linux build
https://answers.unity.com/questions/1086458/local-documentation-for-linux-editor.html

I also searched in /opt/Unity - nothing like Documentation there.

Issue #306, opened 2017-12-14
Use local documentation from current Unity instance


citizenmatt:
The plugin should use the path of the currently running Unity Editor instance to find local documentation.

The current implementation for displaying external documentation of Unity symbols (see #304) will first look up local documentation in the default install locations for Windows, Mac and Linux, before falling back to perform a search on Unity's website. But Unity might not be installed at the default locations, or the currently running editor instance might be a different version installed at a different location. Rider and ReSharper should use the install path of the current Editor instance to find the local documentation.

Unity are going to add `EditorApplication.applicationPath` to the `EditorInstance.json` file (adding the `app_path` property). This gives us the path to the editor executable (`.app` on Mac) but would need manipulating to get to the docs. Hopefully, they'll also add `EditorApplication.applicationContentsPath` too, which will give us the equivalent of (depending on install location):

* `C:\Program Files\Unity\Editor\Data` on Windows
* `/Applications/Unity/Unity.app/Contents` on Mac
* `/opt/Unity-2017.3.0b2/Unity/Editor/Data` or `~/Unity-2017.3.0b2/Unity/Engine/Data` on Linux

Appending `Documentation/en` to this path will give us the local documentation root on all platforms.

van800:
You may update search path logic to similar https://github.com/JetBrains/resharper-unity/blob/master/resharper/src/resharper-unity/Rider/UnityPathTemplateParameter.cs#L34

citizenmatt:
Make sure to also support documentation installed with Unity Hub, as per 3980f917d86312a2c89bd9dc15a28801898aad03:

* `/Applications/Unity/Hub/Editor/2018.1.0/Documentation/en`
* `C:\Program Files\Unity\Hub\Editor\2017.3.1f1\Documentation\en`

And update the online search to include the version:

* `https:////docs.unity3d.com/2017.4/Documentation/ScriptReference/30_search.html?q=...`

rob8861:
@citizenmatt @van800 I am glad this is being worked on. However, to date I have not been able to find a way to open the unity docs directly from Rider, even the online ones will be great. Is there a way from Rider to point the carter at a Unity symbol and open the Unity docs (not the quick docs)???

citizenmatt:
[...]

PR #309, opened 2017-12-19
 add app_path and app_content_path to EditorInstance.json

van800:
I am a bit confused that 
path to documentation is done both on backend in 
ShowUnityHelp.GetDocumentationRoot()
and on frontend in
UnityDocumentationProvider

Thus I would rather not touch it myself.

### DIFF
--- a/resharper/src/resharper-unity/Unity3dRider/Assets/Plugins/Editor/JetBrains/RiderPlugin.cs
+++ b/resharper/src/resharper-unity/Unity3dRider/Assets/Plugins/Editor/JetBrains/RiderPlugin.cs
@@ -279,7 +279,7 @@ namespace Plugins.Editor.JetBrains
   ""process_id"": {0},
   ""version"": ""{1}"",
   ""app_path"": ""{2}"",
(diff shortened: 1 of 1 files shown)


citizenmatt:
This means we're always writing the file now, right? Does the plugin always run after Unity has created the file? Is there any way Unity could create a file over the top of ours?

Would it be better to call the file something different - `RiderEditorInstance.json`, and we change the frontend/backend to look for that file first, and then fall back to `EditorInstance.json`?

I've got #306 to track updating to use the correct information, so unless we want to change the name of the file, I'm happy with this. Approved!

The history of frontend/backend is because ReSharper and Rider have incompatible APIs for this, and the only way I can get it to work for Rider makes it too expensive to check file existence on the backend.

But it could definitely be cleaned up. Perhaps a backend component that reads the file, and pushes the data to the protocol. Then it's easily available for both front and back ends, although the path manipulation will still need to be done on the front end for performance. I'll look at it as part of #306.

And maybe I should chase up plugins being able to extend the protocol...

Issue #323, opened 2018-01-05
Add Error when using Null Coalescence with classes inheriting from UnityEngine.Object


StephenHodgson:
![image](https://user-images.githubusercontent.com/13334553/34622262-5674f4ae-f201-11e7-80b8-cf9a5cb79441.png)

https://blogs.unity3d.com/2014/05/16/custom-operator-should-we-keep-it/

citizenmatt:
The big problem here is all of the other `null` related features that ReSharper/Rider support. We don't want to get in the situation where we ping-pong between suggesting using null-coalescing, ternary or null conditional operators and showing a warning that this code is invalid.

But this is on the radar as something that would be really useful to fix.

See also #35 and #148

Zod-:
I've spent some time digging through the code to find the relevant classes. What I've found is `MergeConditionalExpressionWarning`, which contains the highlight and tooltip, `ConditionalExpressionToAccessFix` provides the quick fix and `ConditionalExpressionAccessAnalyzer` does the analyzing to create the highlight.

I don't know how to actually replace/extend them yet. The only thing I found on this topic is [this](http://www.jetbrains.com/help/resharper/sdk/Features/Actions.html#redefining-action-handlers) outdated snippet in the documentation.

Zod-:
Ok I managed to override the analyzer and prevent the quick fix from showing up. Looks like all you need to do is inherit the analyzer and define it with the same element and highlight types like so:

```cs
    [ElementProblemAnalyzer(typeof(IConditionalTernaryExpression), typeof(MergeConditionalExpressionWarning))]
    public class UnityObjectConditionalExpressionAccessAnalyzer : ConditionalExpressionAccessAnalyzer
    {
        protected override void Run(IConditionalTernaryExpression element, ElementProblemAnalyzerData data, IHighlightingConsumer consumer)
        {
            //base.Run(element, data, consumer);
        }
    }
```

We then can check our stuff and otherwise run the old analyzer with `base.Run`

Zod-:
I'm wondering if this is the right way. There should be a registrar somewhere that contains all the analyzers like for the [quickfixes](https://github.com/JetBrains/resharper-unity/blob/4c3fd20248bf793fc0290e8b91502ca99273f73c/resharper/src/resharper-unity/Feature/Services/QuickFixes/QuickFixRegistrar.cs), right? I think it's cleaner to explicitly replace the analyzer than magically override it through interhitance.

[...]

Issue #148, opened 2017-05-09
Null checking works differently for classes derived from UnityEngine.Object


SirIntruder:
Resharper reports 'this != null' as an error, but really it shouldn't if class is derived from UnityEngine.Object.
This happens often when you work with callbacks, something like

```
public class MyBehaviour : MonoBehaviour
{
  void Start()
  {
    MyInvoker.InvokeAfterSecond(() => 
    {
       // scene gets changed in the meantime, MyBehaviour is now gone 
       // transform getter will throw NPE if this doesn't get checked
        if (this == null) return; 
        transform.position = x;
    });
  }
}
```
btw if somebody knows of a nicer way of checking this do tell me. No unity property (as in gameObject getter) seems to work , they all throw NPE.

nkAlex:
Yeah, this is one of Unity's peculiar behaviours, it can happen when a `GameObject` has been destroyed by Unity (in native code), but the `MonoBehaviour` script is still referenced somewhere and therefore hasn't been garbage collected. In this case none of Unity's built-in properties and methods will work throwing a `NullReferenceException`, and a `this == null` check is the only way to determine if an object is dead or alive (it actually calls to the native code to check the object) and is totally valid for Unity, though makes no sense for C# :)

CAD97:
Here's an old blog post on the issue (couldn't find a newer source): https://blogs.unity3d.com/en/2014/05/16/custom-operator-should-we-keep-it/

citizenmatt:
Rider 2017.3 no longer suggests to convert to `??` or `?.`, and Rider 2018.1 will have a warning if you try and use these operators on a type deriving from `UnityEngine.Object`.

Issue #35, opened 2016-05-21
Supress "Replace '?:'-operator with '??'" suggestion


witcher112:
Hello!

As Unity explains:

> When you do this in Unity:
> if (myGameObject == null) {}
> 
> It behaves inconsistently with the ?? operator, which also does a null check, but that one does a pure c# null check, and cannot be bypassed to call our custom null check.

_Source:_ [http://blogs.unity3d.com/2014/05/16/custom-operator-should-we-keep-it](http://blogs.unity3d.com/2014/05/16/custom-operator-should-we-keep-it/)

This suggestion might mislead the programmer while comparing UnityEngine.Objects.
Would it be possible to supress it?

Thanks in advance!

PS. Huge thanks for this amazing utility :smile: !

citizenmatt:
Rider 2017.3 no longer suggests to convert to `??` or `?.`, and Rider 2018.1 will have a warning if you try and use these operators on a type deriving from `UnityEngine.Object`.

PR #342, opened 2018-01-30
Add null coalescing and propagation analyzer

Zod-:
Relevant: #323, #35, #148 

I would like some input on several sections of the code here where I was unsure. I'll add some review comments to them.

I have not added a quick fix for null propagation as I've found it too complex. Furthermore I have not updated the readme yet to include these features.

Tell me if you want any changes.


resharper/src/resharper-unity/Daemon/Stages/Analysis/UnityNullCoalescingProblemAnalyzer.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            if (expression.LeftOperand == null || expression.RightOperand == null)
+                return;
+            if (!(expression.LeftOperand is IReferenceExpression leftOperand) || !IsDescendantOfUnityObject(leftOperand))
+                return;
+
+            consumer.AddHighlighting(new UnityNullCoalescingWarning(expression));
+        }
+
+        private static bool IsDescendantOfUnityObject([CanBeNull]IReferenceExpression expression)
```
Zod-: I was trying to use `Api.IsUnityType` here but didn't seem to get it to work and ended up implementing my own version. Specifically the check `if (type.Module is IProjectPsiModule projectPsiModule)` fails.
Also the same method exists in `UnityNullPropagationProblemAnalyzer.cs` so if we keep it it should move to a central location.

resharper/src/resharper-unity/Daemon/Stages/Highlightings/UnityNullCoalescingWarning.cs
```diff
@@ -0,0 +1,41 @@
+﻿using JetBrains.DocumentModel;
+using JetBrains.ReSharper.Feature.Services.Daemon;
+using JetBrains.ReSharper.Plugins.Unity.Daemon.Stages.Highlightings;
+using JetBrains.ReSharper.Psi.CSharp;
+using JetBrains.ReSharper.Psi.CSharp.Tree;
+
+[assembly: RegisterConfigurableSeverity(UnityNullCoalescingWarning.HIGHLIGHTING_ID,
+    null, UnityHighlightingGroupIds.Unity, UnityNullCoalescingWarning.MESSAGE,
+    "Unity Object types can't be compared using null coalescing. Use ternary operator instead.",
+    Severity.ERROR)]
```
Zod-: Should the warnings be errors or warnings?

IvanPashchenko: Error is for something that breaks the compilation and Warning is for something that is highly likely to be incorrect.

StephenHodgson: Error should suffice in this case, I believe. Doesn't break compilation, but it always breaks at runtime.

Zod-: I'm pretty sure marking the highlighting as `Error` will actually prevent compilation. Unity doesn't care about it and will just keep compiling but visual studio won't let you build it until you fix the "error".

citizenmatt: No, ReSharper/Rider highlights don't affect compilation. This should be a warning, as it's still valid code, but doing something you most likely didn't intend.

StephenHodgson: That makes sense.

citizenmatt: Yep.

SugoiDev:
Does the "is" matching also internally use null-checks that cause problem with `UnityEngine.Object` types?

Like, 
```
var hashCode = 0;
if (someInstance is UnityEngine.Object unityObject) {
    hashCode = unityObject.GetHashCode();
}
```

Would this be the same as using `hashCode = (someInstance as UnityEngine.Object)?.GetHashCode() ?? 0;`?


If it is, then this is also a place to show a warning when dealing `UnityEngine.Object` derived types.

Zod-:
@SugoiDev Yes, it shows the same behavior.

 |  | GameObject | Destroyed GameObject | Null |
|--------------------------|------------|----------------------|------|
| go == null | ✘ | ✔ | ✔ |
| go is null | ✘ | ✘ | ✔ |
| go is UnityEngine.Object | ✔ | ✔ | ✘ |

citizenmatt:
I'm not sure I understand that table.

I was going to say that we should leave this. The intent of the `is` operator is to convert, and the `if` statement is checking that the conversion worked correctly.

What we're trying to guard against is where the intent is to check the lifetime of the Unity object, but the wrong `null` checks are used (coalescing and propagating operators).

I think these use cases are different enough to warrant leaving it.

Zod-:
The point was that the conversion check succeeds so subsequently you would convert and use it but it's not sufficient enough to be sure it's not destroyed. I agree that this is a different case tho.

citizenmatt:
Very cool PR, thanks! I'll look at merging it tomorrow, and see if we can use the existing `UnityApi` methods.

The only real question I have is with the output of the null coalescing quick fix. It's converting something like:

```
return o ?? this.gameObject;
```

to

```
return o ? o : this.gameObject;
```

Is this the pattern we want to settle on as the "official" guidance as it were? As in, what should be the standard check that ReSharper/Rider prefers? As I see it, we have three choices. Explicit `null` comparison:

```
return o != null ? o : this.gameObject;
```

Or implicit bool comparison:

```
return o ? o : this.gameObject;

// or always default to false to draw more attention to the implicit conversion?
return !o ? this.gameObject : o;
```

Or explicit bool comparison:

```
return o == true ? o : this.gameObject;
```

I suspect that the explicit `null` comparison is the most widely used pattern, but I'm not sure it's the most obvious - it's too easily confused with a normal `null` comparison. An implicit or explicit comparison to `bool` is sufficiently different enough that the intent is more explicit - you're checking lifetime of the unmanaged object, not `null`-ness of the managed object.

This would actually be a nice thing to add into Code Style settings, so the user could choose what style they want (like they can with `var` handling), and we'd underline violations of this and automatically quick fix to the real thing, and we'd just have to pick a default. But annoyingly, the C# CodeStyle page isn't customisable...

resharper/src/resharper-unity/Daemon/Stages/Analysis/UnityNullCoalescingProblemAnalyzer.cs
citizenmatt: This is because the `UnityApi` class is designed to contain the messages/event functions that a class derived from a Unity type can implement, rather than describe all of the Unity API. I think it would be worth renaming that method to be something like `IsDerivedFromUnityType`, or `IsUserUnityType`.

What you've got looks ok, but you could also do [something like this](https://github.com/JetBrains/resharper-unity/blob/4c3fd20248bf793fc0290e8b91502ca99273f73c/resharper/src/resharper-unity/Daemon/Stages/Analysis/IncorrectObjectInstantiationProblemAnalyzer.cs#L29-L33), creating an instance of `IType` for the known name, and using that to pass into a call to `Type.IsSubtypeOf`.

But don't worry about fixing this up now, I want to get this merged. I might have a little refactor of `UnityApi` and add this in there, or something.

Zod-: Yeah, that example is what I was looking for. It's quite hard to understand the relations between all the different types. I still don't really get it. Anyway I gave it a quick test and it works. 

```cs
        protected override void Analyze(INullCoalescingExpression expression, ElementProblemAnalyzerData data, IHighlightingConsumer consumer)
        {
            if (!(expression.LeftOperand is IReferenceExpression leftOperand) || expression.RightOperand == null)
                return;
            var resolve = leftOperand.Reference.Resolve();
            if (resolve.ResolveErrorType != ResolveErrorType.OK)
                return;
            var unityObjectType = TypeFactory.CreateTypeByCLRName(KnownTypes.Object, expression.GetPsiModule());
            if (!leftOperand.Type().IsSubtypeOf(unityObjectType))
                return;

            consumer.AddHighlighting(new UnityNullCoalescingWarning(expression));
        }
```

Unless you want to refactor it later anyway I could commit this to remove some of the redundancy if you want.

citizenmatt: If you've got the commit ready, fire away! 👍

citizenmatt: Yes, the type system interfaces can be confusing. There are two hierarchies in play (with `ITreeNode` mixed in there, too), and they both serve different purposes, and the both have a LOT of derived interfaces.

The `IDeclaredElement` interface provides a language agnostic base for "things that are declared". This can be types, like classes (`ITypeElement`), or variables (`ILocalVariable`), or namespaces (`INamespace`) or file system paths (`IPathDeclaredElement`) or even colours (`IColorDeclaredElement`). But these derived interfaces represent the declarations, and aren't good at representing usages (e.g. `MyClass[]` or `List<string>`).

For that, we have another interface hierarchy, which is a full semantic model. This hierarchy is language specific, and the CLR implementation is `IType` and its base class `IExpressionType`. This can represent all type usages, including arrays (`IArrayType`) and pointers (`IPointerType`). The JavaScript type system has a different, but similar hierarchy, rooted at `IJavaScriptType`.

You can bounce between the two hierarchies - the `IType` hierarchy allows getting back to declared elements (if there are any) via `IDeclaredType.GetTypeElement()`, while the `IDeclaredElement` hierarchy can get to the semantic usage model via `ITypeOwner.Type`.

I usually have to look at base and deriving classes to remember which hierarchy I'm in :)

There's some more detail in the devGuide about [declared elements](https://www.jetbrains.com/help/resharper/sdk/PSI/DeclaredElements.html) and [type systems](https://www.jetbrains.com/help/resharper/sdk/PSI/TypeSystems2.html).

citizenmatt:
Another test: is this something that you'd want to do:

```
if (this == null) return;
```

Rider currently marks that as an always false expression. I'm guessing we'd want that highlight hidden.

citizenmatt:
Awesome stuff. Thanks for the PR, it's very cool to be able to have something to help with Unity's "interesting" `null` handling!

Zod-:
Yeah, checking `this` for null is possible even long after calling `Destroy` until you clear your own references I assume. The objects will be marked as destroyed and unity stuff is emptied out but you can still call your own methods and fields. Interestingly the visual studio debugger tells me the expression is false.
```cs
public class NullTest : MonoBehaviour
{
    public bool IsNull()
    {
        return this == null;
    }
}

    private void Start()
    {
        Destroy(nullTest); //or Destroy(nullTest.gameObject);
        Debug.Log(nullTest.IsNull());
    }
    private void Update()
    {
        Debug.Log(nullTest.IsNull());
    }
```
```txt
False
True
True
True
...
```

citizenmatt:
BTW, we've already branched for 2018.1 EAP1, so this will make it into EAP2.

citizenmatt:
See also [RIDER-14181](https://youtrack.jetbrains.com/issue/RIDER-14181)

PR #346, opened 2018-01-30
ShaderLab culture fixes

Zod-:
This PR fixes 2 section of the ShaderLab that did not properly handle culture variants with doubles when parsing and printing them. e.g. germany uses reversed dot notation (1,000.00) so `0.3` was parsed to `3` and made `ShaderLab.Psi.Colors.ShaderLabColorHighlightingTests.*` fail.

Furthermore `ShaderLab.Psi.Colors.ShaderLabColorPickerTests.TestChangeColor` had the reverse problem where doubles were formatted to `0,7` instead of `0.7`.

On top of the 2 fixes I also added culture specific tests so they run under both english and german.


Zod-:
As for how this bug manifests in the currently released plugin:
Color highlights don't show up for colors that are composed of values other than `1` or `0`. Furthermore when choosing a color with the color picker it will use commas instead of dots as described above.

PR #362, opened 2018-02-15
WIP: Find usages and completion for ShaderLab properties

citizenmatt:
Added a resolve step for ShaderLab, enabling find usages and highlighting usages of ShaderLab properties.

Note that this only works for usages of properties in ShaderLab blocks, and not for usages of the same property as a parameters to Cg/HLSL programs.

- [x] Declared elements for ShaderLab property declarations
- [x] First class references for variable usage in ShaderLab commands (e.g. `Blend [_Src] [_Dst]`)
- [x] Report unresolved property errors
- [x] Report multiple candidates as an error
- [x] Find usages of properties within ShaderLab commands (NOT Cg/HLSL)
- [x] Context highlighting of properties
- [x] Code completion for known properties
- [x] TESTS!


resharper/src/resharper-unity/ShaderLab/Psi/DeclaredElements/ShaderLabDeclaredElementType.cs
```diff
@@ -0,0 +1,27 @@
+﻿using JetBrains.ReSharper.Psi;
+using JetBrains.ReSharper.Psi.Resources;
+using JetBrains.UI.Icons;
+
+namespace JetBrains.ReSharper.Plugins.Unity.ShaderLab.Psi.DeclaredElements
+{
+    public class ShaderLabDeclaredElementType : DeclaredElementType
+    {
+        // TODO: Proper icon
```
IvanPashchenko: Do we want some unity-specific icon here?

resharper/src/resharper-unity/ShaderLab/Psi/DeclaredElements/ShaderLabDeclaredElementPresenter.cs
```diff
@@ -7,19 +9,118 @@ namespace JetBrains.ReSharper.Plugins.Unity.ShaderLab.Psi.DeclaredElements
...
+            }
+
+            if (style.ShowName != NameStyle.NONE && style.ShowName != NameStyle.SHORT &&
+                style.ShowName != NameStyle.SHORT_RAW)
+            {
+                AppendDisplayName(result, element);
+            }
+
+            if (style.ShowConstantValue)
+            {
```
IvanPashchenko: Is this empty code block intentional?

resharper/src/resharper-unity/ShaderLab/Psi/Resolve/IVariableReferenceReference.cs
```diff
@@ -0,0 +1,9 @@
+﻿using JetBrains.ReSharper.Psi.Resolve;
+
+namespace JetBrains.ReSharper.Plugins.Unity.ShaderLab.Psi.Resolve
+{
+    // TODO: Terrible name...
```
IvanPashchenko: It can use a sample of the entity it is describing in the comment then :)
Also, WDYT something like IReferenceToVariableReference (doesn't sound good to me too though)?

resharper/src/resharper-unity/ShaderLab/Psi/DeclaredElements/ShaderLabDeclaredElementType.cs
citizenmatt: Yep. I'll add a request

resharper/src/resharper-unity/ShaderLab/Psi/DeclaredElements/ShaderLabDeclaredElementPresenter.cs
citizenmatt: Not really. I meant to implement it, and forgot :) No-one's using it yet, though. I'll fix it.

PR #378, opened 2018-02-26
fix plugin installer

van800:
Do not do anything about old plugin of 2 files.
Handle only Unity3dRider.cs and new dll format.


resharper/src/resharper-unity/Rider/UnityPluginDetector.cs
```diff
@@ -17,11 +15,9 @@ public class UnityPluginDetector
         
         private readonly ISolution mySolution;
         private readonly ILogger myLogger;
-        private static readonly string[] ourPluginFilesV180 = {"RiderAssetPostprocessor.cs", "RiderPlugin.cs"};
```
IvanPashchenko: Am I right that the last time we used this was pre-release Rider, so there were no stable Rider versions with this plugin variation bundled?

resharper/src/resharper-unity/Rider/UnityPluginDetector.cs
```diff
@@ -193,69 +129,23 @@ private InstallationInfo ExistingInstallation(List<FileSystemPath> pluginFiles)
...
-            }
             
             var pluginDir = parentDirs[0];
-            var filenames = pluginFiles.Select(f => f.Name).ToList();
 
-            // v1.9.0+
             if (pluginFiles.Count == 1)
             {
-                if (pluginFiles[0].Name == MergedPluginFile)
+                if (pluginFiles[0].Name == PluginDllFile)
```
IvanPashchenko: It looks like we won't be able to upgrade from `Unity3DRider.cs` to `.dll` this way.

resharper/src/resharper-unity/Rider/UnityPluginDetector.cs
```diff
@@ -193,69 +129,23 @@ private InstallationInfo ExistingInstallation(List<FileSystemPath> pluginFiles)
...
                 {
-                    var version = GetVersionFromFile(pluginFiles[0]);
+                    var version = new Version(FileVersionInfo.GetVersionInfo(pluginFiles[0].FullPath).FileVersion);
                     return new InstallationInfo(version != ZeroVersion, pluginDir, pluginFiles, version);
                 }
                 
                 myLogger.Warn("One file found, but filename is not the same as v1.9.0+");
```
IvanPashchenko: This error message is obsolete.

resharper/src/resharper-unity/Rider/UnityPluginDetector.cs
van800: yes

Issue #402, opened 2018-03-02
Saving a script in Rider 2018.1 EAP with Rider plugin v2.1.3.2067 triggers a recompile in Unity


SugoiDev:
Using latest Rider 2018.1 EAP with Unity 2018.1.0b9
Rider plugin is v2.1.3.2067

I just let Rider install the latest integration and noticed this behavior.
I have "Auto Refresh" disabled in Unity, but every time I save a script in Rider, I see Unity recompiling in my second monitor.

This isn't very good, since a lot of times when we're coding, it isn't ready to be compiled until we're done.

Removing the integration plugin makes the problem go away.

van800:
I will add "Auto Refresh" as a precondition for that "Refresh on each Save".

SugoiDev:
Sounds good.

PR #406, opened 2018-03-02
Ensure we only sync the solution when needed.

ArtOfSettling:
You are currently forcing a sync everytime the user clicks on a cs file or a log entry. This seems fine, but doesn't happen in Visual Studio, i've tried to mitigate that with these changes, and you should now, only sync the project when needed. (Provided I've caught all the cases)

This is currently, when a cs file has been created, deleted or renamed.
It might be worthwhile considering other situations in the future (assembdefs, rsp, etc)

### DIFF
--- a/unity/JetBrains.Rider.Unity.Editor/EditorPlugin/ModificationPostProcessor.cs
+++ b/unity/JetBrains.Rider.Unity.Editor/EditorPlugin/ModificationPostProcessor.cs
@@ -1,0 +1,36 @@
+using UnityEditor;
+
+namespace JetBrains.Rider.Unity.Editor
+{
+  public class ModificationPostProcessor : UnityEditor.AssetModificationProcessor
+  {
+    public const string ModifiedSource = "com.jetbrains.rider.modifiedsourcefile";
+    
+    private static void OnWillCreateAsset(string path)
+    {
+      var isCs = path.EndsWith(".cs.meta");
+      if (isCs)
+        EditorPrefs.SetBool(ModifiedSource, true);
+    }
+
+    private static AssetDeleteResult OnWillDeleteAsset(string assetPath, RemoveAssetOptions options)
+    {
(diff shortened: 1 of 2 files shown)


van800:
It is great to get a PRs! Thank you!
Just let me know what is the problem with more frequent syncs? Does it affect performance?

WellFiredDevelopment:
Our project is quite large and the sync increases the wait time.

For us, with constant sync the wait time is about 5-6 seconds.

With ‘clever’ sync it’s instant most of the time.

Hope it’s useful for you guys!

van800:
refered to #402

SugoiDev:
Can confirm what @WellFiredDevelopment is experiencing.
The sync delay is noticeable on my current (medium-sized) project.

When a new script file is added using Unity's project view, it will automatically recompile, so the sync part should only be needed if you added a file using an external file manager. In that case, I prefer to ask Unity to recompile myself instead of having it automatic.

I've completely disabled sync to avoid all this.

ArtOfSettling:
To explain a bit more the change I've made, everytime Unity detects an asset was created (through an external source, such as Finder/Explorer/Files or inside Unitys project view), moved, or deleted and that file was a .cs file, it sets a flag that they project should be sync'd.

When rider tries to open a file, it will sync if that flag was set, rather than the current behaviour which sync's all the time

Due to Unity potentially causing a domain reload, I set that flag in Prefs, which to be honest I don't really like. However it's a lightweight approach to solving the problem.

I've been using the script for a little while and it seems to work well, of course, more intensive testing is required.

van800:
Thanks for clarification.
Here is what I meant by asking about AutoRefresh
https://github.com/JetBrains/resharper-unity/commit/16632d0e87c948e1f715ccc6e385f21e5b736972
I narrowed the shouldSync condition to 
(modifiedSource && !EditorApplication.isPlaying && EditorPrefsWrapper.AutoRefresh)

Does it make sense to you?

PR #412, opened 2018-03-03
Multiple features in one branch

van800:
NewtonSoft.JSON reference
SyncSolution
Soft-wraps in log view


citizenmatt:
Why is `newtonsoft.json.dll` checked into the repo instead of being a nuget reference? It would be better not to commit binaries if we can avoid it.

van800:
First I did as a PackageReference and I needed to repack it into the plugin dll, so I added
<CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
so that newtonsoft.dll and other package references are copied to bin folder,
which in turn caused very weird non-stable only local compile errors.

PR #416, opened 2018-03-06
Deterministic builds for the Unity editor plugin

citizenmatt:
The new SDK style .csproj files do not like having multiple projects in the same folder - the file that `dotnet restore` writes to the `obj\` folder with dependency information is overwritten for each project. Last write wins, which is non-deterministic. This can introduce weird build breaks, especially on repeated builds.

This PR adds a `Directory.Build.props` file, like in the `resharper-unity` solution, that will set up bin and obj folders that are per-project, under a `build` folder at the same level as `src`. It also fixes up some of the references that were causing issues - `JetBrains.Annotations.dll` is no longer referenced, as it conflicts with the annotations defined in `UnityEngine.dll`. The `Newtonsoft.Json` reference is now a package reference again, as `dotnet restore` won't overwrite other projects' data files.


citizenmatt:
I think we can probably get rid of `lib\Newtonsoft.Json.dll` now, as I think this should get copied to the output folder correctly now.

It would be nice to get rid of `lib\JetBrains.Annotations.dll`, too. We could make that a package reference, or even better, can we recompile the `Rd*.dll` assemblies so that they don't reference annotations at all? We wouldn't need to reference, or ILRepack then. They must have been compiled with `JETBRAINS_ANNOTATIONS` defined. @korifey?

Issue #422, opened 2018-03-08
Fix up .psi files to prevent generation of Obsolete attributes


citizenmatt:
It's just way too much noise. Disabling gets rid of the noise, but there's a danger we'll miss a proper use of the attribute. We should see if we can fix up the .psi files (or change the generator)

PR #423, opened 2018-03-08
[WIP] Clean up and speed up build

citizenmatt:
**WORK IN PROGRESS. DO NOT MERGE**

This PR is cleaning up and speeding up the build. Specifically:

- [x] Remove all warnings
  - [x] Hide `[Obsolete]` warnings for `TreeNodeVisitor` (#422 is logged for a proper fix)
  - [x] Fix warnings about conflicting VS assemblies
  - [x] Use Rider's own syntax highlighting classes, removing warnings about conflicting class names
  - [x] Remove warnings for importing `ILRepacks.targets` more than once
- [ ] Speed up msbuild - it still takes about 14 seconds to build when nothing has changed (this might be a .NET Core SDK issue. .NET 2.1 is boasting about massive improvements here)
  - [x] Make ErrorsGen support incremental build - causes full recompile due to incorrect inputs in task
  - [x] Make ILRepack support incremental build
  - [x] Make `rdgen` support incremental build - up to date check cache is overwritten, meaning it's always generated, so we always compile
- [x] Improve building Rider plugin/improve gradle build
  - [x] EditorPlugin msbuild step uses same build configuration as rest of script, means no more unnecessary rebuild when building `resharper-unity.sln` if different
  - [x] Assembly version now comes purely from msbuild properties, passed from gradle script. No more generating `AssemblyInfo.cs` or modifying `Packaging.props`. Reduces unnecessary rebuilds
  - [x] Refactor to use task types for repeated tasks (e.g. dotnet restore, msbuild, nuget pack)
  - [x] Pass local Rider SDK version as a property instead of patching the .props file
  - [x] Remove patching of plugin.xml - the IntelliJ plugin is already doing that for us
  - [x] Create super fast path for building and running Rider with plugin
- [x] Create shared run configuration to allow running Rider with plugin directly from Rider
- [x] Improve Travis output
  - [x] Disable fancy gradle output formatting
  - [x] Initialise gradle wrapper before we start the actual build, separate the output
  - [x] Folding blocks of output

Stretch goals :)

- :x: Create shared run configuration to allow _debugging_ Rider with plugin directly from Rider - This requires a custom run configuration
- :x: Create shared run configuration to allow running/debugging ReSharper from Rider - didn't investigate
- :x: Add schema to the errors.xml files - didn't investigate
- :x: Ensure PsiGen works correctly with just the Rider project. See [RIDER-14366](https://youtrack.jetbrains.com/issue/RIDER-14366) - No changes to existing implementation

Notable differences:

* EditorPlugin will be built in Release mode. Was defaulting to Debug
* Gradle script won't do a rebuild if only the version number changes. I don't think this will cause a problem as we release from CI, and that will rebuild
* Removes the commented out stale nuget cache cleaner
* Removes the `SinceBuild` and `UntilBuild` parameters - they're not used. Can easily be added back and set in `patchPluginXml` task

### DIFF
--- a/resharper/src/.idea/.idea.resharper-unity/.idea/runConfigurations/runIde.xml
+++ b/resharper/src/.idea/.idea.resharper-unity/.idea/runConfigurations/runIde.xml
@@ -1,0 +1,17 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="runIde" type="DotNetProject" factoryName=".NET Project">
+    <option name="EXE_PATH" value="$PROJECT_DIR$/GradleLauncher/bin/Debug/GradleLauncher.exe" />
+    <option name="PROGRAM_PARAMETERS" value="../../../../../rider/ runIde -P dotNetUpToDate=true" />
+    <option name="WORKING_DIRECTORY" value="$PROJECT_DIR$/GradleLauncher/bin/Debug" />
+    <option name="PASS_PARENT_ENVS" value="1" />
+    <option name="USE_EXTERNAL_CONSOLE" value="0" />
+    <option name="USE_MONO" value="0" />
+    <option name="PROJECT_PATH" value="$PROJECT_DIR$/GradleLauncher/GradleLauncher.csproj" />
+    <option name="PROJECT_EXE_PATH_TRACKING" value="1" />
+    <option name="PROJECT_ARGUMENTS_TRACKING" value="1" />
+    <option name="PROJECT_WORKING_DIRECTORY_TRACKING" value="1" />
+    <option name="PROJECT_KIND" value="Console" />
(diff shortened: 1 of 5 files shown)


rider/backend.gradle
```diff
@@ -5,15 +5,99 @@ import org.apache.tools.ant.taskdefs.condition.Os
...
+    @TaskAction
+    def restore() {
+        if (project.SkipNuGetRestore) {
+            logger.lifecycle('SkipNuGetRestore=true, skipping restore')
+            return
+        }
+
+        project.backend.ReportProgress("Restoring packages for $packagesFile")
+
+        // TODO: What is Source?
```
IvanPashchenko: `Source` is always set on CI and it is the directory which contains the Rider SDKs that the plugin is being built against.

PR #434, opened 2018-03-13
Wave12 skr unity2

kskrygan:
Minimized UI for Unity solutions
Discover UnityTest
Launch UnityTest and NUnit through UnityEditor


IvanPashchenko:
Do not merge, waiting for green tests with bundled SDK on TC.

Issue #387, opened 2018-02-28
EditorPlugin is not installed


KasperKnop:
According to [this](https://github.com/JetBrains/resharper-unity/issues/266) post the issue was resolved in 2018.1, but in the new 2018.1 EAP I still cannot use multiple solutions at the same time?

van800:
Have you cleaned up Unity3dRider.cs?

KasperKnop:
How do I do that exactly? I deleted the script and now I cannot get Rider to reinstall it in my project.

van800:
New plugin is named JetBrains.Rider.Unity.Editor.Plugin.Repacked.dll

van800:
If still plugin is not installed, please zip and attach all log files from Help->Show log in Files.

KasperKnop:
My Rider settings have always been set up to install and update the Rider Editor plugin automatically. I have this issue on both my work laptop and home PC. Here are the [log](https://github.com/JetBrains/resharper-unity/files/1770376/log.zip) files from my laptop.

van800:
Very soon (maybe tomorrow or after weekend) nightly builds with fixed plugin installation will become available. Can we wait that new build, if it will not work for you, we will make a call and investigate it together. Is it fine for you?

KasperKnop:
Sure thing! Thanks :)

van800:
Nightly build is already available.
It is accessible via https://www.jetbrains.com/toolbox/download/

KasperKnop:
Thanks, but the plugin still does not install itself on either of my two computers with the nightly build.

Here is the [log.zip](https://github.com/JetBrains/resharper-unity/files/1777360/log.zip)

van800:
Logs say
11:09:42.899 |I| UnityPluginDetector           | No unity generated projects found. Skipping installation.
11:09:42.899 |I| UnityPluginInstaller          | Plugin should not be installed.

Is it true that you work with Unity-generated solution?

KasperKnop:
I'm not sure I understand your question.

If I create a new project in Unity, Unity will create my Microsoft Visual Studio Solution file.

When I then create a new script and open it with Rider as my default editor, no plugin get's added to my Unity project.

van800:
Lets have a call on Monday-Tuesday. I am really curious, what is a problem. Do you work Europe time?
Skype or google meeting? Thanks!

KasperKnop:
[...]

PR #436, opened 2018-03-13
fix installing plugin when TargetFramework version generated by Unity…

van800:
… is not good and reference to UnityEngine is unresolved


van800:
https://github.com/JetBrains/resharper-unity/issues/387

PR #449, opened 2018-03-15
Improve heuristics for detecting Unity-generated project on frontend

van800:
Check Assets folder and UnityEngine reference to hide Nuget toolwindow.


rider/src/main/kotlin/com/jetbrains/rider/UnityReferenceDiscoverer.kt
```diff
@@ -34,10 +35,18 @@ class UnityReferenceDiscoverer(project: Project) : LifetimedProjectComponent(pro
...
             isUnityProject = true
+            isUnityGeneratedProject = hasAssetsFolder(project)
         }
     }
 
     fun addUnityReferenceListener(listener: UnityReferenceListener) {
         myEventDispatcher.addListener(listener)
     }
+
+    companion object {
```
kskrygan: let's make global extension function instead.

PR #450, opened 2018-03-15
Feature/unity explorer1

xtmq:


citizenmatt:
LGTM, but a couple of questions:

1. Why are we showing references in this view? Isn't this about file system/assets, rather than projects?
1. Is the references node under the root? If so, which project does it apply to? There could be a lot of projects under the tree, including .asmdef based projects
1. Shouldn't the icons also have `.png` files? I thought the icon loader wold strip the `.svg` and look for `@2x.png` and `.png` before then actually rendering the `.svg`? Has that changed?

Issue #350, opened 2018-02-01
Autodetect Unity mono using info in /Library/EditorInstance.json


van800:
Most of the times our Unity plugin writes a special file to Solution
SlnDir/Library/EditorInstance.json
{
 "process_id": 13136,
 "version": "2017.3.0f2",
 "app_path": "F:/Program Files/Unity2017.3.0f1/Editor/Unity.exe",
 "app_contents_path": "F:/Program Files/Unity2017.3.0f1/Editor/Data"
}

"app_contents_path" may be used to determine mono path.

This is related to https://youtrack.jetbrains.com/issue/RIDER-12319
Implementation via IMonoPathProvider and UnityMonoPathProvider

van800:
Modern Unity versions generate csproj, which require msbuild. But Unity doesn't ship MSBuild.
So autodetecting xbuild doesn't make any sense.

PR #458, opened 2018-03-16
[WIP] Autodetect Unity mono using info in /Library/EditorInstance.json #350

van800:
I don't know how to make it work, since new UnityGeneratedSolutionMonoPathProvider is a solution component, not shell.


van800:
New Unity has roslyn, but doesn't have MSbuild.
So this becomes not really a working way

PR #464, opened 2018-03-19
fix visibility of UnityToolbar for non-Unity projects

kskrygan:


van800:
Why not calling fun isUnityGeneratedProject instead of isUnityProject?

PR #467, opened 2018-03-20
Skr fixes1

kskrygan:
Fix launch Unity Tests for 2018.1


van800:
since there is no references to new api exception reflection calls, you should use UnityVersion instead of ifdefs.

PR #473, opened 2018-03-23
Connection problem notifications

ArsenyChernyaev:
Refactor run and attach configurations;
Add debug output listeners for both cases


ArsenyChernyaev:
Done. Now message from custom debugger output listener should be listed after basic message. Could you please check it again?

van800:
image
Check "Editor Attaching" in Unity settings message should be duplicated to Console, I guess.

PR #486, opened 2018-03-28
Collect statistics

van800:
- [x] Collect statistics of used Unity versions and scripting runtimes
- [ ] Calls for force Refresh in Edit/Play modes separately.
- [ ] Game actions (Play/Pause...)
- [ ] AutoRefresh setting in Unity (enabled/disabled)

Initial request  RIDER-12292

Lets merge what is done and see, what we collect. I will continue this in 2018.2


resharper/src/resharper-unity/Rider/UnityEditorProtocol.cs
```diff
@@ -219,6 +224,18 @@ private void BindPluginPathToSettings(Lifetime lf, EditorPluginModel model)
                     model.FullPluginPath.SetValue(string.Empty);
                 });
         }
+        
+        private void TrackActivity(EditorPluginModel model, Lifetime lf)
+        {
+            if (!model.ApplicationVersion.HasValue())
```
van800: I am not sure about this piece. It looks weird. Any idea how to improve it?

Issue #490, opened 2018-04-01
Incremental compiler alpha triggers rebuild on any code change


JayPavlina:
Unity released an [alpha for its incremental compiler](https://forum.unity.com/threads/unity-incremental-c-compiler.523993/). It mostly works, but every time I change something, Rider starts reanalyzing the whole project. Rider also doesn't seem to detect C# 7.2 support.

Just wanted to make you aware of this in case its something you want to work on supporting.

citizenmatt:
I think in order to get 7.2 support, you need to make sure that you've got .net 4 selected in Unity. Do you have a repro project we can take a look at to see about the reanalysing? Also, what kind of analysis is it doing - indexing or the full solution wide analysis?

JayPavlina:
I have .net 4.x selected. Somebody in the thread [mentioned the same thing happening](https://forum.unity.com/threads/unity-incremental-c-compiler.523993/page-2#post-3445052) with the reanalysis. It's reanalyzing the whole solution. You can probably test with any project.

SugoiDev:
Also noticed that issue on my side.

I'm grateful that rider doesn't block the UI and close my files randomly when it reloads the projects/re-analyze the solution, like VS. This is actually what drove me away from VS in the first place.

But having the analyzer resetting on every tiny recompile isn't ideal.

The incremental compiler creates one .rsp for each project. Maybe Rider will be able to tap to that and prevent full reloads of the projects. Instead, it can update the projects model internally using the data from those response files.

citizenmatt:
It would be great if anyone has a test project ready to go, otherwise this will have to wait until I have chance to take a look at that page and create a new project.

SugoiDev:
If @JayPavlina doesn't have one, I will attach one later.

citizenmatt:
Great, thanks!

van800:
@JayPavlina Are you on Windows? Do you have dotnet 4.7.1 installed? Please check if you have it in
`C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework`

van800:
I have fixed potential problem with rewriting LangVersion multiple times, but I am not completely sure that it was the case. I guess it will become public in nightly builds in couple of days.

But It would be still good to have a repro solution + video-recording of the problem.
@JayPavlina @SugoiDev

SugoiDev:
With the incremental compiler, setting the language version should become unnecessary, since the compiler is already doing them.

[...]

PR #491, opened 2018-04-02
Only set csproj properties if value is higher

citizenmatt:
Makes sure we only set `TargetFrameworkVersion` or `LangVersion` if the version we're setting is higher than what's already there. So if a newer Unity (such as the incremental compiler) sets the framework version to 4.7.1, we won't override it with 4.6. Same with `LangVersion`

### DIFF
--- a/unity/JetBrains.Rider.Unity.Editor/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
+++ b/unity/JetBrains.Rider.Unity.Editor/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
@@ -7,6 +7,7 @@ using System.Xml.Linq;
 using JetBrains.Util;
 using JetBrains.Util.Logging;
 using UnityEditor;
+using UnityEngine;
 
 namespace JetBrains.Rider.Unity.Editor.AssetPostprocessors
 {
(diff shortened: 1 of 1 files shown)


van800:
If the target pack of dot net framework is not installed?

citizenmatt:
What do you mean?

van800:
In the case, when user has some dotnet installed, previously we were trying
to use it, but with your change we will fail to open solution and complain
about missing dotnet, right?

citizenmatt:
Hmm. Yes, I've changed the behaviour here. Previously, we would set the target framework version to the latest installed .NET Framework version (e.g. 4.7) when targeting .net 4 and the oldest installed .NET Framework (e.g. 3.5) when targeting .net 2. 

But only on Windows. On Mac/Linux, we would use the default of 4.6 or 3.5, unless it was overridden by the user. And this means that if the new incremental compiler sets it to 4.7.1, we'd overwrite it and downgrade it.

So yes, with this change, we let Unity or a higher priority plugin decide the target framework version more often, and so it could get it "wrong". But only if it's a higher version than we would set - is this more likely to be installed?

Is this something that would be properly solved if we always used the Unity Mono toolset? Would all versions of Unity ship with a toolset that we could use to correctly load the projects with the correctly referenced target framework version, so we wouldn't have to change it at all? More importantly, is this possible?

(And there's also a bug in this change that would only write out two parts of the version - 4.7 instead of 4.7.1, which is also wrong.)

van800:
@citizenmatt will it help with this?
https://github.com/JetBrains/resharper-unity/issues/490
Can I take this both problems for myself?

citizenmatt:
It didn't start as a fix for that, but I think would sort out the C# 7.2 bit, yes.

This started because I'd seen someone post to a forum or something that Unity would now set C# and target framework version, but we always overwrite it, and I though it would be better if we only changed it if it was actually wrong or missing, so we don't have to duplicate Unity logic.

Feel free to take this over.

van800:
close in favor of #506

PR #492, opened 2018-04-02
Wave12 extensible protocol

IvanPashchenko:
Summary of changes:
* Frontend-Backend model is now called `UnityModel` and is encapsulated as a component in a `UnityHost`.
* Rider-Unity Editor model is now called `EditorPluginModel`

* Protocol files are located in `rider/protocol/model`
* Gradle task `generateModel` generates both `UnityModel` and `EditorPluginModel`

* `protocol.gradle` file is doing all the proper `rdgen` calls

* no changes made to `customData` approach as this would've made merges impossible, but after merge we can finally refactor our models

P.S.: @van800 : `EditorPlugin.Tests` moved to `xUnit` as I'm tired of dealing with of assembly conflicts between nUnits, sorry :(


.gitignore
```diff
@@ -41,6 +41,7 @@ resharper/src/.idea/**/*.iml
 
 # ContentModel.xml is also auto-generated
 resharper/src/.idea/**/contentModel.xml
+resharper/src/.idea/**/other.xml
```
citizenmatt: Can we stop `other.xml` from being written? @kskrygan 

https://github.com/JetBrains/resharper-unity/blob/master/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/ui/UnityUIManager.kt#L19

resharper/src/resharper-unity/Rider/UnityHost.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace JetBrains.ReSharper.Plugins.Unity.Rider
+{
+    [SolutionComponent]
+    public class UnityHost
+    {
+        public RdUnityModel Model { get; }
+
+        public UnityHost(ISolution solution)
+        {
+            Model = solution.GetProtocolSolution().GetRdUnityModel();
```
kskrygan: Why do we need such a component? The clients could get access to model with smth like
solution.GetUnityModel(), where GetUnityModel() would be an extension method.

IvanPashchenko: It is much more sustainable in a long way when we'll have to add some Unity-specific not related to the protocol. Plus injecting the component seems much more convenient to me than `solution.GetProtocolSolution().GetRdUnityModel();` in each constructor. Encapsulating the protocol just seems more readable to me, I guess, and we have a lot of them in RIder.

Issue #358, opened 2018-02-12
Unity Explorer - csharp script file templates is not available (in the empty folders) 


van800:
UnityExplorer shows all folders, which is good.
But not all the folders are in any of the projects.
So for folders, which are in not in any project it is not easy to add a csharp file.

In Unity Explorer action Add class for folder, which is not in any csproj is not available
![image](https://user-images.githubusercontent.com/1482681/36093267-6cd9d9b0-0fea-11e8-8791-3fded57716f8.png)

van800:
it is not possible to provide a cs file template for folder outside of any project.

van800:
related https://youtrack.jetbrains.com/issue/RIDER-14200

PR #493, opened 2018-04-03
File Templates

xtmq:
#358

### DIFF
--- a/resharper/src/resharper-unity/ProjectExtensions.cs
+++ b/resharper/src/resharper-unity/ProjectExtensions.cs
@@ -1,8 +1,10 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using JetBrains.Annotations;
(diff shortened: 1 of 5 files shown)


resharper/src/resharper-unity/ProjectExtensions.cs
```diff
@@ -28,6 +32,12 @@ public static class ProjectExtensions
             ourUnityEditorReferenceName, ourUnityEngineReferenceName, ourUnityEngineCoreModuleReferenceName, ourUnityEngineSharedInternalsModuleReferenceName
         };
 
+        public static bool IsUnitySolution([NotNull] this ISolution solution)
+        {
```
IvanPashchenko: Maybe an assert here for solution load being fully done?

xtmq: I think 'false' in that case is OK =)

PR #495, opened 2018-04-04
New approach for supporting old Unity versions

van800:
1. We compile two editor-plugin dlls now
2. First one has basic support for all Unity versions
3. If Unity version is >=5.6 First dll will substitute itself with Second one.

known problem
opening in old Unity after Second dll was added is not supported, we consider it an extremely minor case.


resharper/src/resharper-unity/Rider/UnityEditorProtocol.cs
```diff
@@ -32,25 +35,30 @@ public class UnityEditorProtocol
...
         private readonly IShellLocks myLocks;
         private readonly ISolution mySolution;
+        private readonly PluginPathsProvider myPluginPathsProvider;
         private readonly UnityHost myHost;
 
         private readonly ReadonlyToken myReadonlyToken = new ReadonlyToken("unityModelReadonlyToken");
         public readonly Platform.RdFramework.Util.Signal<bool> Refresh = new Platform.RdFramework.Util.Signal<bool>();
         
         private readonly IProperty<EditorPluginModel> myUnityModel;
+        private IContextBoundSettingsStoreLive myBoundSettingsStore;
```
IvanPashchenko: This can be made readonly

resharper/src/resharper-unity/Rider/UnityPluginDetector.cs
```diff
@@ -83,6 +97,21 @@ public InstallationInfo GetInstallationInfo(FileSystemPath previousInstallationD
             }
         }
 
+        private bool TryFindInSolution(ISolution solution, out InstallationInfo result)
+        {
+            result = ShouldNotInstall;
+            var location = solution.GetAllAssemblies()
+                .Select(assembly => assembly.GetLocation())
+                .FirstOrDefault(t => t.Name == PluginPathsProvider.BasicPluginDllFile);
```
IvanPashchenko: What if it's a full file that we need to upgrade?

unity/JetBrains.Rider.Unity.Editor/EditorPlugin/PluginEntryPoint.cs
```diff
@@ -133,18 +133,19 @@ private static void Init()
...
-            riderProtocolController.Wire);
+          var protocol = new Protocol("UnityEditorPlugin", serializers, identities, MainThreadDispatcher.Instance, riderProtocolController.Wire);
           ourLogger.Log(LoggingLevel.VERBOSE, "Create UnityModel and advise for new sessions...");
           var model = new EditorPluginModel(connectionLifetime, protocol);
           AdviseUnityActions(model, connectionLifetime);
           AdviseModel(model);
           OnModelInitialization(new UnityModelAndLifetime(model, connectionLifetime));
           AdviseRefresh(model);
           
+          model.FullPluginPath.Advise(connectionLifetime, AdditionalPluginsInstaller.UpdateSelf);
```
IvanPashchenko: If there isn't a race because we don't set this property until UnityModel is set, maybe we need a comment on both sides, this seems like an easy place to actually create a race. Maybe we should provide a separate property IsEditorPluginReady for that?

PR #506, opened 2018-04-06
Target framework and LangVersion.

van800:
Set our value if it is lower than Unity one.
+ Some small fixes


unity/JetBrains.Rider.Unity.Editor/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
```diff
@@ -332,54 +333,89 @@ private static void ApplyCustomReference(string name, XElement projectContentEle
...
-
-      if (UnityUtils.ScriptingRuntime > 0)
+      SetOrUpdateProperty(projectElement, xmlns, "TargetFrameworkVersion", s =>
         {
-          var version = PluginSettings.TargetFrameworkVersion;
-          if (!PluginSettings.OverrideTargetFrameworkVersion)
+          string version = string.Empty;
+          try
           {
+            version = s.Substring(1);
```
citizenmatt: This will throw an exception if the string is less than 1 char long - so if there's no `TargetFrameworkVersion` element, or it's empty, we'll get an exception.

unity/JetBrains.Rider.Unity.Editor/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
```diff
@@ -332,54 +333,89 @@ private static void ApplyCustomReference(string name, XElement projectContentEle
...
               {
-                version = versions.OrderBy(v => new Version(v)).Last();
+                var versionOrderedList = versions.OrderBy(v1 => new Version(v1));
+                var foundVersion = UnityUtils.ScriptingRuntime > 0
+                  ? versionOrderedList.Last()
+                  : versionOrderedList.First();
+                // roslyn compiler requires dotnet 4.7.1, which may not be present
+                var fvIsParsed = decimal.TryParse(foundVersion, NumberStyles.Number, CultureInfo.InvariantCulture, out var fv);
+                var vIsParsed = decimal.TryParse(version, NumberStyles.Number, CultureInfo.InvariantCulture, out var v);
+                if (fvIsParsed && vIsParsed && fv > v)
```
citizenmatt: Why parse as decimals? `Version` allows comparison

unity/JetBrains.Rider.Unity.Editor/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
```diff
@@ -332,54 +333,89 @@ private static void ApplyCustomReference(string name, XElement projectContentEle
...
+        if (expected == "latest" || existing == "latest")
+          return "latest";
+
+        // Only use our version if it's not already set, or it's less than what we would set
+        // Note that if existing is "default", we'll override it
+        var currentIsParsed = decimal.TryParse(existing, NumberStyles.Number, CultureInfo.InvariantCulture, out var currentLanguageLevel);
+        var expectedIsParsed = decimal.TryParse(expected,NumberStyles.Number, CultureInfo.InvariantCulture, out var expectedLanguageLevel);
+        if (currentIsParsed && expectedIsParsed && currentLanguageLevel < expectedLanguageLevel
+            || !currentIsParsed
+            )
```
citizenmatt: It might be better to use `float` here instead of `decimal`. And perhaps check the return of `TryParse` and set defaults, to make the conditional easier? If you can't parse current, set it to `float.MinValue`, if you can't parse expected, set it to `float.MaxValue`, then it's a simple comparison for all conditions?

unity/JetBrains.Rider.Unity.Editor/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
van800: These are strings. And dotnet 3.5 doesn't have Version.TryParse

unity/JetBrains.Rider.Unity.Editor/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
van800: when debugging floats look weird 3.5 becomes 3.499991 ))

unity/JetBrains.Rider.Unity.Editor/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
van800: it is under try-cath. I added additional check for null or empty.

unity/JetBrains.Rider.Unity.Editor/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
citizenmatt: Just parse as `Version` but inside a try/catch? Parsing as decimal will fail with 4.6.1 and 4.6.2

unity/JetBrains.Rider.Unity.Editor/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
citizenmatt: Yikes! Is that a bug in the debugger or mono?

unity/JetBrains.Rider.Unity.Editor/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
van800: switched to Version

Issue #507, opened 2018-04-06
Clearing log view does not clear details pane


citizenmatt:
If a log entry is selected, and showing the stack trace in the details pane, clicking the clear button will clear the list of log entries, but doesn't clear the details pane. The old stack trace is still visible.

van800:
fixed in latest nightly

PR #508, opened 2018-04-09
2 last fixes for release

van800:
Remove hack with updating nunit-framework.dll
Clearing log view does not clear details pane #507


van800:
@citizenmatt FYI

PR #546, opened 2018-04-30
combobox to run unity tests via nunit

kskrygan:


rider/src/main/kotlin/com/jetbrains/rider/UnityReferenceDiscoverer.kt
```diff
@@ -50,3 +53,18 @@ class UnityReferenceDiscoverer(project: Project) : LifetimedProjectComponent(pro
...
+    val component = this.getComponent<UnityReferenceDiscoverer>()
+    return component.isUnityProject
+}
+
+fun Project.isConnectedToEditor(): Boolean {
+    val component = this.getComponent<UnityHost>()
+    return component.sessionInitialized.value
+}
+
+fun Project.isConnectedToEditorLive(): Property<Boolean> {
```
van800: When Unity connection disappears, I can't see drop-down is changed anyhow.

kskrygan: I was thinking about this, but decided not to advise for such a live property. I don't want this drop down to blink on every autosave (refresh in Unity).

Issue #554, opened 2018-05-09
(Rider) Plugin hides configuration selector


Chippit:
Having this plugin hides the default debug/release platform configuration selector from the toolbar.

I understand that this is not useful to most Unity users, but in our projects we have updated the solution files to change symbol defines and referenced assemblies based on various configuration settings, and now it is not (trivially) possible to switch between them.

Is it possible to make the disabling of this configuration selector an option in the plugin's settings?

van800:
relates to https://github.com/JetBrains/resharper-unity/issues/549

van800:
Option for this was added in the last nightly build
Distributed via https://www.jetbrains.com/toolbox/app/

tbg10101:
Sorry to resurrect an old ticket but the platform confutation option is again missing here in 2019. I want to switch to Debug so that Rider uses the DefineConstants in the csproj.

The Release target has no defines so it is hard to use platform-conditional compilation with Rider.

van800:
@tbg10101 
There is a way to restore "Full UI".
![image](https://user-images.githubusercontent.com/1482681/69491664-60cbac80-0e98-11ea-8224-0027f04d5aaf.png)
Let me know, if it worked for you.

tbg10101:
I don't have that option:

![image](https://user-images.githubusercontent.com/1455209/69504483-26422e00-0ef1-11ea-8e4a-3ff3a4b8142b.png)

van800:
Please attach logs from Help-Compress logs and show in Explorer.

Chippit:
I see the confusion, @tbg10101. You're pressing the wrong button. You want the Unity icon to the left of the Refresh Unity Assets.

![Annotation 2019-11-25 102152](https://user-images.githubusercontent.com/870779/69523839-824a9800-0f6d-11ea-8dfe-f1c9f0aaf2d7.jpg)

van800:
Oh man, you are right! Thanks @Chippit

Issue #549, opened 2018-05-02
Incompatibility with Database plugin


RDeluxe:
I have a weird bug since a while. As soon as I enable the Unity plugin, I can't use the native database plugin anymore.

This manifests in several ways : 

* Can't find the database window under `View -> Windows -> Database` anymore
* Using the command palette to launch the database tools results in "Nothing here"
* The keyboard shortcut mapped to the database tool is doing nothing

If I disable the Unity plugin everything works fine. As soon as I enable it and restart, the database windows is going to disappear soon after launch (I'm guessing when the Unity plugin enables).

Rider version 2018.1
Unity 2018.1

To be clear, the tool is still working, but I can't access it easily => 

A quick  and dirty video : https://bit.ly/2HOFOMl

citizenmatt:
We hide some tool windows when we open a Unity project, by design. We want to simplify the interface for Unity developers, and we've made a start with some tool windows - NuGet is one, the Database window is another. If you open a non-Unity project, the Database tooling should be still available. Do you need to use the Database tooling from a Unity project? /cc @kskrygan

RDeluxe:
Hello.

Yes, I do. I'm developing a multiplayer game and being able to have a quick look on my data is paramount.

It makes sense for this to be a default behaviour, however imvho : 

* It should be clearly documented somewhere
* It should be reversible easily

RAACosta:
👍 To this issue. I don't believe it's a good idea to hide/remove IDE functionalities just because it's a Unity project unless it's completely incompatible with the Unity solution.

Please bring it back.

marcelschmidt1337:
I am also using it with a Unity project to look up player data. I thought it got secretly removed, glad I found this ticket here.

RDeluxe:
Same as you @marcelschmidt1337 , it took me some time to debug this. 
As @RAACosta it's definitely not advised to hide such a basic feature.

I would be glad to have a temporary solution, because it seems I can access the tool only after the first reboot after activating the unity plugin.

marcelschmidt1337:
@RDeluxe I've created a new project with an empty solution and there I use the database tool for now. Hope this is fixed soon.

RDeluxe:
I'm actually using DataGrip myself

van800:
Database functionality was fully restored in the last nightly build.
Distributed via https://www.jetbrains.com/toolbox/app/

PR #555, opened 2018-05-09
Add options to show hide UI parts

van800:
https://github.com/JetBrains/resharper-unity/issues/549
https://github.com/JetBrains/resharper-unity/issues/554


resharper/src/resharper-unity/Rider/UnityOptionsPage.cs
```diff
@@ -28,6 +28,10 @@ public class UnityOptionsPage : OptionsPageBase
 
             CheckBox((UnitySettings s) => s.InstallUnity3DRiderPlugin, "Install or update Rider plugin automatically");
             CheckBox((UnitySettings s) => s.AllowAutomaticRefreshInUnity, "Automatically refresh Assets in Unity");
+            
+            Header("UI");
```
IvanPashchenko: This is going the way of having too many checkboxes. And we're still hiding NuGet window for example. Let's just make one option which configures whether we are in a simplified view or not. @kskrygan

rider/frontend.gradle
```diff
@@ -185,7 +185,7 @@ prepareSandbox {
     dependsOn prepareForBuild
 
     doLast {
-        assert file(unityeditorplugin_path).exists()
+        //assert file(unityeditorplugin_path).exists()
```
IvanPashchenko: is this intended?

resharper/src/resharper-unity/Rider/UnityOptionsPage.cs
van800: If you look at requests, you will see that one person requested "Solution Configuration" and multiple others - "Database". And showing Nuget would be misleading/harmful, because there is absolutely no chance it may work in the generated solution.

rider/frontend.gradle
van800: I need your advice on this, lets talk today.

PR #577, opened 2018-05-31
Modify csproj generation to handle RoslynAnalyzers

van800:
Howto:

Create "RoslynAnalyzers" near sln file
Install nuget or Copy dll with custom Roslyn Analyser. Good examble is
https://www.nuget.org/packages/UnityEngineAnalyzer/
Assets Sync C# Project in Unity
Rider will show custom analysis results
image
Example from https://github.com/meng-hui/UnityEngineAnalyzer

### DIFF
--- a/unity/JetBrains.Rider.Unity.Editor/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
+++ b/unity/JetBrains.Rider.Unity.Editor/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
@@ -67,6 +67,7 @@ namespace JetBrains.Rider.Unity.Editor.AssetPostprocessors
       FixSystemXml(projectContentElement, xmlns);
       SetLangVersion(projectContentElement, xmlns);
       SetProjectFlavour(projectContentElement, xmlns);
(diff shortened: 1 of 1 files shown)


citizenmatt:
I have reservations about this - I don't think Rider should be responsible for creating procedures such as this. If there's an existing standard way for adding Analysers, then we should support it, but if there isn't, we shouldn't invent one.

van800:
@citizenmatt @kskrygan 
I haven't got [any response](https://unity.slack.com/archives/C9BJY3RM2/p1527677402000374) in Unity slack about this.
If we don't want to include it directly to our package, I will publish it as a separate gist useful for anybody interested. Rider/VS/Any_other_editor.

SugoiDev:
With the new Incremental Compiler, there is some support for analyzers, but it isn't documented (and probably not even working yet). I'm not sure what direction they'll take on this, since they've gone kinda silent on the entire Incremental Compiler thing. You might want to try asking Dom_Laflamme about this.


In any case, I think that, even if there is no standard way of adding Roslyn analyzers right now, extending the project generation to be able to have then included is great and might open this path to many.

van800:
New and better aproach https://github.com/vad710/UnityEngineAnalyzer/pull/43
Works for Unity 2019.2+

PR #582, opened 2018-06-06
switch to Full UI

kskrygan:


van800:
Consider user solution in the same folder with unity-generated one.
I guess, it will read the "other.xml" file and show minimized UI.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/ui/UnityUIManager.kt
```diff
@@ -15,15 +16,22 @@ import com.jetbrains.rider.util.lifetime.Lifetime
 import com.jetbrains.rider.util.lifetime.LifetimeDefinition
 import com.jetbrains.rider.util.reactive.Property
 import com.jetbrains.rider.util.reactive.whenTrue
+import org.jdom.Element
 
-@State(name = "UnityProjectConfiguration", storages = [(Storage(StoragePathMacros.WORKSPACE_FILE))])
+@State(name = "UnityProjectConfiguration", storages = [(Storage(value = "other.xml"))])
```
citizenmatt: Does this create a file called `other.xml`? Can we keep it in `workspace.xml` instead? Otherwise it's another file that we need to teach users whether they should add it to source control or not, and the name is really not descriptive.

PR #589, opened 2018-06-11
Invoke rdgen through new Gradle plugin

yole:


citizenmatt:
Nice! 👍 But it looks like the repo is only available behind VPN - this is why the AppVeyor and Travis builds are failing. Are we going to make it publicly available?

Issue #520, opened 2018-04-13
Add option to disable Refresh on focusing Unity during Play mode


van800:
Track Play mode and call EditorApplication.LockReloadAssemblies on entering it
EditorApplication.UnLockReloadAssemblies on exiting it.

Will this be a helpful feature?
Given that it is already supported by https://assetstore.unity.com/packages/tools/utilities/console-enhanced-free-42381
Should we add it too?

van800:
related https://capeguy.co.uk/2015/06/no-more-unity-hot-reload/

lifangjie:
yes, i think it is a good idea.

Nezz:
We are using Console Enhanced at my company and some have experimented with that feature but it was rather unreliable. Unity just stopped compiling at some point no matter what. This was about a year ago though, so things might have improved.

van800:
fixed with https://github.com/JetBrains/resharper-unity/pull/615

PR #609, opened 2018-06-24
Implemented assembly locking

JurjenBiewenga:
This fixes issue #520

### DIFF
--- a/unity/EditorPlugin/PluginEntryPoint.cs
+++ b/unity/EditorPlugin/PluginEntryPoint.cs
@@ -164,7 +164,7 @@ namespace JetBrains.Rider.Unity.Editor
 
       PlayModeStateChanged += (state, newState) =>
       {
-        if (PluginSettings.PreventAssemblyReloadDuringPlay)
+        if (PluginSettings.AssemblyReloadSettings == AssemblyReloadSettings.Delayed)
         {
           if (newState == PlayModeState.Playing)
           {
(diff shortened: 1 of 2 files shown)


van800:
Hi! 
EditorGUILayout.Popup would be better UX than just Toggle. Consider the following:
"Compile during Play"
 - "Immediate"
 - "Manual"
We may later add something like "Compile on Stop", "Stop and Recompile" options.

JurjenBiewenga:
Hey, thanks for the feedback.
I'll change it around to popup and I'll add stop and recompile as well.

van800:
Hey! I finally had time for review.
Few problems here.
1. It all doesn't work when there is no Rider protocol connection. Just leave Unity and close Rider - 
riderProtocolController.Wire.Connected.WhenTrue will never fire.
2. New names are not self-explanatory. But don't focus on that much, we can change them later.
3. Doing constant checks on       EditorApplication.update += () =>
is not a good idea, I guess.
Try using  AppDomain.CurrentDomain.DomainUnload += (sender, args) =>
instead.

IvanPashchenko:
@JurjenBiewenga hey, thanks for the PR!

Have you submitted a contributor license agreement? If not, please follow the steps described at http://www.jetbrains.com/agreements/cla/ to sign it.

@citizenmatt we should probably say something about CLA in the `readme.md`. Or set up an `contributors.md`.

van800:
Oh man, I am a bit sad for the time we have spent on this. Occasionally I found out that exactly same feature was added by Unity team in Unity 2018.2. 
image

JurjenBiewenga:
Arguably it would still be a good feature as a lot of users are stuck on the 2017 or even the 5.x versions of Unity.

van800:
I did some testing, everything seems working.
I added two tiny finish changes in branch
https://github.com/JetBrains/resharper-unity/commits/pr-609
If @JurjenBiewenga you agree, I will merge it.

JurjenBiewenga:
Thanks for the changes. Feel free to merge it.

PR #612, opened 2018-06-26
Simple text filter for LogView

van800:


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/toolWindow/log/LogSmartSearchField.kt
```diff
@@ -0,0 +1,42 @@
...
+    init {
+        RiderUI.setHeight(this, 25)
+        minimumSize = Dimension(JBUI.scale(100), minimumSize.height)
+
+        font = RiderUI.BigFont
+        textEditor.emptyText.isShowAboveCenter = true
+
+        RiderUI.overrideKeyStroke(textEditor, "shift ENTER", { transferFocusBackward() })
+        RiderUI.overrideKeyStroke(textEditor, "ENTER", { goToList() })
+        //RiderNuGetPrereleaseHelper.overrideKeyStroke(viewModel, textEditor)
```
IvanPashchenko: do we need all this commented code?

PR #615, opened 2018-06-27
Pr 609 Implemented assembly locking during play

van800:
Discussed in https://github.com/JetBrains/resharper-unity/pull/609


van800:
@JurjenBiewenga I made one more change for initial editor state in the correctly, could you please look it through? Am I dropping some cases out?

van800:
What is the purpose of reloading on Pause?

PR #629, opened 2018-07-06
[WIP] Add Packages node to Unity Explorer

citizenmatt:
- [x] Show "Packages" node if the project supports Packages
- [x] Use display name and show folder name, version and description in tooltip
- [x] Truncate the list of project names shown next to the folder name to 3 projects. Show the rest in a tooltip
- [x] Show "local" packages, where the files don't live under `Packages` but are listed in `manifest.json` with a `file:` URL.
- [x] Show modules + read-only referenced packages
- [x] Show dependencies not explicitly listed in `manifest.json`
- [x] Sort the tree alphabetically on display name
- [x] Update the node if the folder structure of any of the packages changes
- [x] Locate in solution explorer for all types of packages
- [x] Resolve module information against the manifests in app-specific `BuiltInPackages` folder
- [x] Correct paths for Linux
- [x] Refresh tree when `manifest.json` is edited
- [x] Refresh package metadata when an embedded package's `package.json` is edited
- [ ] Proper icons for Packages node + package folders
- [x] Error handling and logging


citizenmatt:
@xtmq Can you take a look at this please? I'm not sure I'm reading the json files in the right place. Also, I'm not sure on the best way to handle updates to a) the list of folders, b) modifications to the package.json metadata.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/explorer/PackagesRoot.kt
```diff
@@ -0,0 +1,103 @@
...
+class PackagesRoot(project: Project, virtualFile: VirtualFile)
+    : UnityExplorerNode(project, virtualFile, listOf()) {
+
+    override fun update(presentation: PresentationData) {
+        if (!virtualFile.isValid) return
+        presentation.presentableText = "Packages"
+        presentation.setIcon(UnityIcons.Explorer.PackagesRoot)
+    }
+
+    // TODO: This doesn't work if there are two nodes. Doesn't even get called
```
xtmq: It will be fixed in IDEA 18..3 (at least I was told so)

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/explorer/UnityExplorerNode.kt
```diff
@@ -123,17 +145,18 @@ open class UnityExplorerNode(project: Project,
...
-    class Root(project: Project, virtualFile: VirtualFile)
+    class AssetsRoot(project: Project, virtualFile: VirtualFile)
         : UnityExplorerNode(project, virtualFile, listOf()) {
 
         private val referenceRoot = ReferenceRoot(project)
 
         override fun update(presentation: PresentationData) {
             if (!virtualFile.isValid) return
-            presentation.presentableText = project!!.solutionName
+            presentation.presentableText = "Assets"
```
xtmq: Why 'Assets' instead of solution name?

citizenmatt: Because it describes what's there - this node shows Assets, the other node shows Packages. It was a bit weird having "GemShader" and "Packages". And the project name is available in the title bar, etc.

xtmq: But if there is no 'Packages' folder, the root node 'Assets' also looks strange... at least for me

PR #646, opened 2018-07-13
Improve "attach and play" run configurations

citizenmatt:
Changes the current "Attach and play" from a separate run configuration type to an additional factory for the existing "Attach" type. Should be merged before 2018.2, as it changes the serialised IDs of the new configuration type.

### DIFF
--- a/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/util/UnityIcons.kt
+++ b/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/util/UnityIcons.kt
@@ -12,12 +12,6 @@ class UnityIcons {
 
             // TODO: Proper icons!
             @JvmField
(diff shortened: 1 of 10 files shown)


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/util/UnityIcons.kt
```diff
@@ -100,6 +94,17 @@ class UnityIcons {
...
             val OpenEditorLog = FilterEditModeMessages
             val OpenPlayerLog = FilterPlayModeMessages
+
+            val AttachToUnity = Icons.UnityLogo
+        }
+    }
+
+    class RunConfigurations {
+        companion object {
+            // TODO: Proper icons!
```
IvanPashchenko: Are these proper icons? Do we have an issue on youtrack for them?

PR #656, opened 2018-07-18
Stop plugin being updated on each project load

citizenmatt:
The plugin is currently being updated every time the project is loaded.

* The file version of the plugin wasn't being set, so was always `0.0.0.0`
* The installer was checking against a "zero" version before installing. But the zero version was `0.0` which doesn't match `0.0.0.0`, so it was always installed.

New version:

* Always installs if there is no existing plugin
* Conditionally installs if the new version is different to the existing version
* Always installs if the new version ends with `.9998` or `.9999`, which is what we use while debugging. This allows for testing new builds of the plugin without requiring a version change


van800:
I am on vacation till end of July, I will not have an opportunity to make a
review earlier, sorry.

PR #683, opened 2018-08-06
182 disable handle package file conflicts

van800:


van800:
It seems I pushed some extra. I will make a new clean PR

PR #690, opened 2018-08-07
Support csc.rsp (used instead of mcs.rsp)

van800:
1. Avoid showing UnityPackageManager process in the list of processes to attach with default Attach to Unity Editor config.
2. support csc.rsp the same way as mcs.rsp was supported


citizenmatt:
LGTM. Why swapping to portable pdb? Does it work with ILRepack?

van800:
Switching to portable pdbs seems to change really nothing, so I reverted it back.

PR #693, opened 2018-08-07
fix docs being incorrect about EditorWindow.OnProjectChange

### LINKED ISSUE #679: OnProjectChange event of Unity's EditorWindow class is recognized as "never used" in last version. It was working fine before.
OnProjectChange event of Unity's EditorWindow class is recognized as "never used" in last version. It was working fine before.

IvanPashchenko:
fix ApiParser to work with hub on windows
fixes #679


apiParser/src/Program.cs
```diff
@@ -34,6 +34,7 @@ public static class Program
```
IvanPashchenko: @citizenmatt should I do something with that list?

citizenmatt: Nah. It's useful if you need to completely recreate the docs list, but you need ALL of those folders installed somewhere, and that's several gigabytes of html. It's much easier just to incrementally update it by comparing the existing API list against a new docs folder. Leave this around if we ever have a need to completely rebuild. I know we probably won't, and it's in source control and everything, but, I can't get too worked up about an internal tool :)

apiParser/src/Program.cs
```diff
@@ -228,6 +231,15 @@ private static void FixDataFromIncorrectDocs(UnityApi unityApi)
                 foreach (var function in type.FindEventFunctions("OnAssignMaterialModel"))
                     function.SetReturnType(new ApiType("UnityEngine.Material"));
             }
+
```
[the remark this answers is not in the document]
IvanPashchenko: done

PR #701, opened 2018-08-09
Disable error for invalid asmdef reference

### LINKED ISSUE #647: Rider 2018.2 : Errors with Assembly Definition Files and Packages folder
Rider 2018.2 EAP
Build #RD-182.3458.490

When using any Unity Packages dependencies (such as Unity.Entities) at any .asmdef in your projects, rider shows as "Unknown project reference".

image

May be related to #476 or #629

citizenmatt:
Temporarily disable the error for invalid asmdef references, because it can't resolve against assembly definitions defined in files external to the project. A more permanent fix is to include those files into a custom PSI module. I'll raise another issue to do that post 2018.2

Fixes #647. Fixes RIDER-17700


DavidAntliff:
@citizenmatt "A more permanent fix is to include those files into a custom PSI module. I'll raise another issue to do that post 2018.2" - do you know which issue, if any, covers this case?

citizenmatt:
The issue is #702, but still hasn't been addressed.

PR #720, opened 2018-08-20
add entities component data to the list of object fields/properties

ArsenyChernyaev:


resharper/src/resharper-unity/Rider/Debugger/UnityComponentData/UnityComponentDataSource.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        public IDebuggerHierarchicalObject ParentSource { get; }
+        public string Name => "Component Data";
+        public SoftEvaluationContext Context { get; }
+
+        public ObjectValue[] GetChildren(ObjectPath path, int index, int count, IEvaluationOptions options)
+        {
+            try
+            {
+                var allocatorTempObject =
```
IvanPashchenko: would be nice to extract stuff under try into separate methods

PR #722, opened 2018-08-22
disable dotcover only for solutions, which can establish protocol con…

van800:
…nection to Unity


resharper/src/resharper-unity/ProjectModel/UnitySolutionHostSyncListener.cs
```diff
@@ -0,0 +1,37 @@
...
+    public class UnitySolutionHostSyncListener:SolutionHostSyncListener
+    {
+        private readonly ISolution mySolution;
+
+        public UnitySolutionHostSyncListener(ISolution solution)
+        {
+            mySolution = solution;
+        }
+
+        private bool? isAbleToEstablishProtocolConnectionWithUnityCached;
```
IvanPashchenko: `Lazy<bool>`?

PR #729, opened 2018-08-24
Version check typo fix

jorgeamado:
SetManuallyDefinedCompilerSettings - must be called for unity version < 2017
In existing code if unity version is less then 2017 - this code will be skipped

Please accept asap
Tarkov developer will be very happy )


van800:
Thanks a lot.

Issue #721, opened 2018-08-22
Using a dynamic requires Microsoft.CSharp.dll


PandawanFr:
When I use the (.Net 4.6) `dynamic` type and try to use the `Equals` method using that dynamic object, ReSharper gives me the error _"One or more types required to compile a dynamic expression cannot be found. Are you missing references to Microsoft.CSharp.dll and System.Core.dll?"_

The only way to fix it was to add Microsoft.CSharp.dll to references.

I am guessing this is ReSharper thinking that Microsoft.CSharp.dll isn't included while it probably has a different name/equivalent in Unity (which is why it doesn't return that error inside Unity). So I'm thinking that maybe `resharper-unity` could add support for dynamics in some way so that it doesn't do that error. 

Here's a picture of the error itself. (In this example, `GetValue()` returns a dynamic type).
![Imgur](https://i.imgur.com/ZE9e3Qh.png)

SugoiDev:
Got that too quite a while ago, and ended up just dropping a copy of Microsoft.CSharp.dll on my project.
I only use it on the editor, so I don't know if it has any side effect in a build (or if it works at all).

PandawanFr:
Yeah, that's what I ended up doing as well. But I'm worried that it might have effects on the build in some way. Also, it'd be nice not to export "useless" packages.

van800:
Please try mcs.rsp as suggested here and let me know the results.
https://youtrack.jetbrains.com/issue/RIDER-17277#focus=streamItem-27-3001597-0-0

SugoiDev:
This workaround doesn't work with the new Incremental Compiler.
Creating a csc.rsp with the reference to Microsoft.CSharp will show the warning below:

---

C:\Users\User\AppData\Local\Unity\cache\packages\staging-packages.unity.com\com.unity.incrementalcompiler@0.0.42-preview.19\.bin\Microsoft.CSharp.dll(0,0): warning MSB3245: Could not resolve this reference. Could not locate the assembly "C:\Users\User\AppData\Local\Unity\cache\packages\staging-packages.unity.com\com.unity.incrementalcompiler@0.0.42-preview.19\.bin\Microsoft.CSharp.dll". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors.

van800:
@SugoiDev I believe it is a Unity issue.
@citizenmatt how do you think should we implement a workaround or accept a PR on this?
[...]

PR #740, opened 2018-08-27
add reference to Microsoft.CSharp.dll for new mono runtime and NET 4.x…

van800:
… compatibility level

fixes https://github.com/JetBrains/resharper-unity/issues/721


unity/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
```diff
@@ -120,13 +120,27 @@ private static bool UpgradeProjectFile(string projectFile, XDocument doc)
...
       changed |= SetManuallyDefinedCompilerSettings(projectFile, projectContentElement, xmlns);
+      changed |= SetAdditionalReference("Microsoft.CSharp.dll", projectContentElement, xmlns);
       changed |= SetXCodeDllReference("UnityEditor.iOS.Extensions.Xcode.dll", projectContentElement, xmlns);
       changed |= SetXCodeDllReference("UnityEditor.iOS.Extensions.Common.dll", projectContentElement, xmlns);
       changed |= SetDisableHandlePackageFileConflicts(projectContentElement, xmlns);
 
       return changed;
     }
 
+    private static bool SetAdditionalReference(string referenceName, XElement projectContentElement, XNamespace xmlns)
```
citizenmatt: It would be better to call this something like `AddMicrosoftCSharpReference` and don't bother passing the reference name, because the method already makes assumptions that it's .net 4.x only, so it's not a general purpose method.

unity/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
```diff
@@ -353,32 +369,39 @@ private static bool ApplyManualCompilerSettingsReferences(XElement projectConten
...
-              hintPath = dllPath.FullName;
-          }
-        }
+    [CanBeNull]
+    private static string GetHintPath(string name)
+    {
+      string hintPath = null;
 
-        ApplyCustomReference(name, projectContentElement, xmlns, hintPath);
+      if (PluginSettings.SystemInfoRiderPlugin.operatingSystemFamily == OperatingSystemFamilyRider.Windows)
```
citizenmatt: Why is this only for Windows? Why don't we do the same for Mac/Linux? It works without hint path on Mac, so what's the issue we're solving for Windows?

van800: On Windows MSBuild will 100% resolve it from dotnet framework targets path 
On Mac/Linux in case when xbuild from Unity is used - it will resolve it from Unity reference assemblies.
However you are right about the case, when non-Unity MsBuild is used. I will add hintpaths for all os.

van800:
Added hintpaths for all OS

PR #746, opened 2018-08-30
skip ReSpeller in asmdef files #RIDER-17701

van800:


resharper/src/resharper-unity/Rider/ReSpeller/UnityByExtensionFileFilter.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace JetBrains.ReSharper.Plugins.Unity.Rider.ReSpeller
+{
+    [Language(typeof (KnownLanguage))]
+    public class UnityByExtensionFileFilter : ITypoAnalyzerFileFilter
+    {
+        public bool ShouldSkipFile(IFile file)
+        {
+            var sourceFile = file.GetSourceFile();
+            var path = sourceFile != null ? sourceFile.ToProjectFile()?.Location : null;
```
IvanPashchenko: Artem advised to take `GetLocation` from source file instead of doing `ToProjectFile`

citizenmatt:
Uh-oh. This is a breaking API change. This should be a `183` thing. Won't affect Rider since we're bundled, but it will definitely break the ReSharper plugin - we'll be implementing an interface that doesn't exist in 2018.2 RTM.

IvanPashchenko:
@citizenmatt it is building though (and we are using proper R# SDK from intellij-repo). Maybe we should fixate 182 RTM as a branch?

IvanPashchenko:
we want this fix in 2018.2.2

IvanPashchenko:
We already have a separate rtm branch, so this merge won’t affect R# plugin if we build from here

citizenmatt:
It will build, because we always build against the latest SDK, which is the maintenance release one. But we shouldn't be changing APIs in releases - it _always_ causes problems for plugin devs. 

If we really, really must have it in 2018.2.2, then `#ifdef RIDER`. We can't take it in the R# plugin. (But I'm still cross about a breaking API change in a maintenance release)

IvanPashchenko:
Ah, I understand now. Agreed.

Issue #751, opened 2018-08-30
Unity plugin from Rider 2018.2.1 doesn't set LangVersion correctly


renattou:
After updating to Rider 2018.2.1 (plugin version 2018.2.0.78), the Unity plugin sets the LangVersion tag as empty. Like this: `<LangVersion></LangVersion>`.
This causes Rider to suggest C#6 features even though we use C#4.
Downgrading to 2018.1.4 (plugin version 2018.1.0.133) fix the problem, setting it as `<LangVersion>4</LangVersion>`.

van800:
@renattou I can't reproduce it.
What is your Unity version and OS?

Make sure you have "Override LangVersion" unchecked
![image](https://user-images.githubusercontent.com/1482681/44881757-08b3f380-acb1-11e8-93f9-d57766d7d3fd.png)

Please set LoggingLevel to Verbose and call Assets->Sync C# Project, logs will appear in %TEMP%\Unity3dRider or $TMDDIR\Unity3dRider depending on the OS. Thanks!

renattou:
@van800 I'm using using Unity 5.6.5p3 on macOS 10.13.6.

Using Rider 2018.1.4: [2018.1.4.log](https://github.com/JetBrains/resharper-unity/files/2338155/2018.1.4.log)
Using Rider 2018.2.1: [2018.2.1.log](https://github.com/JetBrains/resharper-unity/files/2338156/2018.2.1.log)

I checked the override option, but it was already unchecked.

renattou:
I found this comment https://gist.github.com/van800/875ce55eaf88d65b105d010d7b38a8d4#gistcomment-2689159
which could be associated with this issue.

van800:
@renattou Does it work if you check "Override LangVersion" and specifically change it to 4?

van800:
I was able to reproduce the problem. Thank you.

van800:
Until build with fix is out, you may use "Override LangVersion" setting.

renattou:
@van800 glad to hear you could reproduce it! I'm using the override setting for now and it's working. Thanks!

van800:
Fix is available in the nightly builds, which are accessible via https://www.jetbrains.com/toolbox/app/

PR #752, opened 2018-08-31
Set LangVersion in case when Property itself was missing #751

van800:


unity/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
```diff
@@ -509,7 +509,8 @@ private static bool SetLangVersion(XElement projectElement, XNamespace xmlns)
         // Note that if existing is "default", we'll override it
         var currentIsParsed = VersionExtensions.TryParse(existing, out var currentLanguageLevel);
         var expectedIsParsed = VersionExtensions.TryParse(expected, out var expectedLanguageLevel);
-        if (currentIsParsed && expectedIsParsed && currentLanguageLevel < expectedLanguageLevel)
+        if (currentIsParsed && expectedIsParsed && currentLanguageLevel < expectedLanguageLevel
+            || !currentIsParsed)
```
citizenmatt: Presumably this fails because `LangVersion` isn't even set in the file (I think later versions of Unity started writing this themselves), so the incoming `existing` value will be `string.Empty`. We don't handle empty strings very well, here or in `SetOrUpdateProperty`.

I think we should be more explicit about our intentions when we get an empty string - if it's unset, always set it to `expected`:

```
if (string.IsNullOrEmpty(existing))
  return expected;
```

And then just leave this `if` statement as it was.

Finally, we should also change `SetOrUpdateProperty` to not write or update anything if the returned value is `string.Empty` -  this would stop us getting empty elements. Perhaps with some logging.

van800: Adding Property with empty value was never a case, but it may still be valid.

citizenmatt: Yeah, that comment was wrong before any of these changes 😊 What do you think we should do for `"default"`? Just leave it, like we do with `"latest"`, or replace it with `expected`?

PR #753, opened 2018-08-31
Optimize plugin startup performance

van800:
Avoid using JsonConvert.SerializeObject which is slow, use StringBuilder instead.

### DIFF
--- a/unity/EditorPlugin/NonUnity/RiderProtocolController.cs
+++ b/unity/EditorPlugin/NonUnity/RiderProtocolController.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
+using System.Text;
 using JetBrains.DataFlow;
 using JetBrains.Platform.RdFramework;
 using JetBrains.Platform.RdFramework.Impl;
@@ -32,7 +34,7 @@ namespace JetBrains.Rider.Unity.Editor.NonUnity
     }
   }
   
-  [Serializable]
+//  [Serializable]
   class ProtocolInstance
   {
     public int Port;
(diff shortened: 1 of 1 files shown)


unity/EditorPlugin/NonUnity/RiderProtocolController.cs
```diff
@@ -46,7 +48,15 @@ public ProtocolInstance(int port, string solutionName)
...
+        //return JsonConvert.SerializeObject(connections); //turns out to be slow https://github.com/JetBrains/resharper-unity/issues/728 
+      var sb = new StringBuilder("[");
+
+      sb.Append(connections
+        .Select(connection=> "{" + $"\"Port\":{connection.Port},\"SolutionName\":\"{connection.SolutionName}\"" + "}")
+        .Aggregate((a, b) => a + "," + b));
+
+      sb.Append("]");
+      return sb.ToString();
     }
```
citizenmatt: A `foreach` might be even better here - LINQ can be quite expensive.

citizenmatt:
Does this mean we can get rid of the JSON.NET reference? That will make our plugin smaller, which should make it load (a tiny bit) faster.

unity/EditorPlugin/NonUnity/RiderProtocolController.cs
van800: I don't like removing trailing comma, Aggregate is safer. I can hardly imagine someone having more than 3-4 solution files in this folder.

van800:
There is one more usage https://github.com/JetBrains/resharper-unity/blob/182/unity/EditorPlugin/RiderPathLocator.cs#L290

citizenmatt:
Ah. What defines do we normally have set? Both projects in the solution have `UNITY_2017_3` set - do we ever build against an earlier version of UnityEngine.dll?

van800:
It should be a problem. I will check some old Unity version.
We do build in `unity\JetBrains.Rider.Unity.Editor.sln`, but it is for validation purposes only.

van800:
Yeah, it brakes Unity 4.7 and 5.0
Could not load type 'UnityEngine.JsonUtility' from assembly 'JetBrains.Rider.Unity.Editor.Plugin.Repacked'.
  at JetBrains.Rider.Unity.Editor.RiderPathLocator+<>c__DisplayClass13_0.<CollectPathsFromToolbox>b__1 (System.String a) [0x00000] in <filename unknown>:0

van800:
Added fix https://github.com/JetBrains/resharper-unity/pull/762

PR #762, opened 2018-09-04
fix EditorPlugin for Unity 4.7 and 5.0, which don't have JsonUtility

van800:


StephenHodgson:
Surprised to see you guys are still supporting back to `UNITY_4_7`! Cool tho.

citizenmatt:
I think 4.7 is "best effort" support :) Our API knowledge (code completion, etc.) starts with Unity 5.0

van800:
I can't figure out how to remove defines without dropping support to Unity versions prior to 5.4. 
Consider also that we provide RiderPathLocator to Unity team to facilitate getting all installed Riders in Unity 2018.3+.
Are we ok to merge this fix? @citizenmatt

citizenmatt:
Yep, LGTM.

But I am confused about what we use the defines for. As I understand it, the plugin we actually release is compiled against Unity 2017.3 (both the one we install for < 5.6 and the full version), while the 4.7 and 5.5 projects are just for validating that the plugin still compiles for those versions and we aren't using any APIs that we shouldn't be. But if we're using defines, we're compiling different APIs, and the extra projects aren't helping us catch incorrect API usage. So should we compile all projects with no defines and the same APIs on all versions, and then we know it compiles and runs on all versions of Unity?

van800:
Ok, then this would be logical 
https://github.com/JetBrains/resharper-unity/pull/766
Reference libs from 4.7 and having similar Define.
I am sorry, I made a mess of two PR-s instead of one.

PR #763, opened 2018-09-04
QF: Use explicit type instead of string in some methods

krasnotsvetov:
Performance Issue: Using the string for the component type variants results in additional overhead that should not be incurred when possible. 

How to fix it: Instead use explicit type like so, `GetComponent<TextMesh>();`


resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/TypeInsteadOfStringAnalyzer.cs
```diff
@@ -0,0 +1,98 @@
...
+using JetBrains.ReSharper.Psi.CSharp.Util.Literals;
+using JetBrains.ReSharper.Psi.Resolve;
+using JetBrains.ReSharper.Psi.Tree;
+using JetBrains.ReSharper.Psi.Util;
+
+namespace JetBrains.ReSharper.Plugins.Unity.CSharp.Daemon.Stages.Analysis
+{
+    [ElementProblemAnalyzer(typeof(IInvocationExpression), HighlightingTypes =
+        new[] {typeof(TypeInsteadOfStringUsingWarning), typeof(InvalidStringForTypeWarning)})]
+    public class TypeInsteadStringAnalyze : UnityElementProblemAnalyzer<IInvocationExpression>
```
IvanPashchenko: `UseExplicitTypeInsteadOfString` would be better IMO. 
@citizenmatt WDYT?

IvanPashchenko:
@krasnotsvetov merge conflict is present

resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/TypeInsteadOfStringAnalyzer.cs
krasnotsvetov: I think it's better too, so I renamed file.

resharper/src/resharper-unity/CSharp/Daemon/Errors/CSharpErrors.xml
```diff
@@ -180,7 +192,34 @@
...
+      <Argument>AvailableTypes.Length > 1 ? "\nThe type defined in several namespaces and behavior is unexpected" : string.Empty</Argument>
+    </Message> 
+    <Range>InvocationMethod.GetHighlightingRange()</Range>
+    <Behavour overlapResolvePolicy="WARNING" />
+  </Warning>
 
+  <Warning name="InvalidStringForType" configurableSeverity="Unity.InvalidStringForType">
+    <Parameter type="ILiteralExpression" name="argument" />
+    <Parameter type="string" name="typeLiteral" />
+    <Message value="{0} could not be found or any type with same name is unsuitable here.">
```
IvanPashchenko: let's distinguish between this cases
and tell why it is unsuitable

resharper/src/resharper-unity/CSharp/Daemon/Errors/CSharpErrors.xml
```diff
@@ -180,7 +192,34 @@
...
+    <Parameter type="IInvocationExpression" name="invocationMethod" />
+    <Parameter type="string" name="methodName" isValid="skip"/>
+    <Parameter type="string" name="stringLiteral" isValid="skip"/>
+    <Parameter type="IExpression" name="argument" />
+    <Parameter type="ITypeElement[]" name="availableTypes" isValid="skip"/>
+    <Message value="Using the string for the component type variants results in additional overhead and is not safe. 
+    Suggestion is transform from .{0}(\&quot;{1}\&quot;) to .{0}&lt;{1}&gt;().{2}.">
+      <Argument>MethodName</Argument>
+      <Argument>StringLiteral</Argument>
+      <Argument>AvailableTypes.Length > 1 ? "\nThe type defined in several namespaces and behavior is unexpected" : string.Empty</Argument>
```
IvanPashchenko: separate warnings

resharper/src/resharper-unity/CSharp/Daemon/Errors/CSharpErrors.xml
IvanPashchenko: Also let's surrond the class name with quotes here, e.g.
`"Whatever" could not be found`

resharper/src/resharper-unity/CSharp/Feature/Services/QuickFixes/UseExplicitTypeInsteadOfStringQuickFix.cs
```diff
@@ -0,0 +1,138 @@
...
+                        var qualifier = referenceExpression.QualifierExpression;
+                        if (qualifier != null)
+                        {
+                            builder.Argument(qualifier);
+                            builder.Append(".");
+                        }
+
+                        builder.Append(myMethodName);
+                        builder.Append("<");
+                        builder.Append(myType.GetClrName().FullName);
```
TessenR:
```cs
builder.Argument(myMethodName);
...
builder.Argument(TypeFactory.CreateType(myType));
```
will be safer here.

resharper/src/resharper-unity/CSharp/Feature/Services/QuickFixes/UseExplicitTypeInsteadOfStringQuickFix.cs
```diff
@@ -0,0 +1,138 @@
...
+                        builder.Append(">");
+                        builder.Append("()");
+
+                        var newInvocation = factory.CreateExpression(builder.ToString(), builder.ToArguments());
+                        var result = ModificationUtil.ReplaceChild(myOldInvocation, newInvocation);
+
+                        // beatify generated reference
+                        var resultTextRange = result.GetDocumentRange();
```
TessenR: This part should be redundant if you use `builder.Argument(IType);` instead of a fully qualified name.
Current implementation also unnecessarily optimizes references in qualifier which might not be expected. Might worth adding a test.

Also please check `.IsValid()` when using document ranges.

resharper/src/resharper-unity/CSharp/Feature/Services/QuickFixes/UseExplicitTypeInsteadOfStringQuickFix.cs
```diff
@@ -0,0 +1,138 @@
...
+                                NullProgressIndicator.Instance);
+                        }
+                    }
+                }
+
+                return null;
+            }
+
+            public override string Text =>
+                $"Change '.{myMethodName}(\"{myLiteralName}\")' to '.{myMethodName}<{myType.GetClrName()}>()'";
```
TessenR: This might generate really long QF text that will unnecessary mention the original code and fully qualified name of the target type.
Wouldn't something like `Use type argument`/`Use type-safe overload` be just as descriptive here?

resharper/src/resharper-unity/CSharp/Feature/Services/QuickFixes/UseExplicitTypeInsteadOfStringQuickFix.cs
```diff
@@ -0,0 +1,138 @@
...
+                        }
+
+                        builder.Append(myMethodName);
+                        builder.Append("<");
+                        builder.Append(myType.GetClrName().FullName);
+                        builder.Append(">");
+                        builder.Append("()");
+
+                        var newInvocation = factory.CreateExpression(builder.ToString(), builder.ToArguments());
+                        var result = ModificationUtil.ReplaceChild(myOldInvocation, newInvocation);
```
TessenR: `var result = myOldInvocation.ReplaceBy(newInvocation);`
You also won't need the containing `using (WriteLockCookie.Create())` this way

resharper/src/resharper-unity/CSharp/Daemon/Errors/CSharpErrors.xml
```diff
@@ -180,6 +204,53 @@
...
+    <Parameter type="string" name="methodName" isValid="skip"/>
+    <Parameter type="string" name="stringLiteral" isValid="skip"/>
+    <Parameter type="IExpression" name="argument" />
+    <Parameter type="ITypeElement[]" name="availableTypes" isValid="skip"/>
+    <Message value="Using the string for the component type variants results in additional overhead and is not safe. 
+    Suggestion is transform from '.{0}(\&quot;{1}\&quot;)' to '.{0}&lt;{1}&gt;()'.">
+      <Argument>MethodName</Argument>
+      <Argument>StringLiteral</Argument>
+    </Message> 
+    <Range>InvocationMethod.GetHighlightingRange()</Range>
```
TessenR: Please consider limiting highlighting ranges for these warnings.
Current highlighting for the whole invocation:
```
var foo = someQualifier.GetUnityObject().GetComponent("Foo");
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
could be limited to just the string literal:
```
var foo = someQualifier.GetUnityObject().GetComponent("Foo");
                                                       ^^^
```
or at least the last part of an invocation:
```
var foo = someQualifier.GetUnityObject().GetComponent("Foo");
                                         ^^^^^^^^^^^^^^^^^^^
```

resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/UseExplicitTypeInsteadOfStringAnalyzer.cs
```diff
@@ -0,0 +1,175 @@
...
+                consumer.AddHighlighting(new InvalidStringForTypeWarning(argument as ILiteralExpression, stringLiteral));
+                return;
+            }
+
+            if (filteredTypes.Length >= 2)
+            {
+                consumer.AddHighlighting(new AmbiguousUnityMonoBehaviourDefinitionWarning(argument as ILiteralExpression, stringLiteral));
+            }
+            
+            consumer.AddHighlighting(new UseExplicitTypeInsteadOfStringUsingWarning(expression, methodName, stringLiteral, argument, filteredTypes));
```
TessenR: Please add a check for preprocessor directives.`if (!expression.ContainsPreprocessorDirectives())`
Current implementation might suggest incorrect changes in cases like this one:
```cs
class Foo : MonoBehavior { }
class Test : MonoBehavior
{
  public object M()
  {
    return GetComponent(
#if DEBUG
    "DebugType"
#else
    "ReleaseType"
#endif
    );
  }
}
```

I believe it's worth supporting preprocessor here as such requests for components look like a convenient point to customize debug/release behavior.

resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/UseExplicitTypeInsteadOfStringAnalyzer.cs
```diff
@@ -0,0 +1,175 @@
...
+            
+            var typeNode = factory.CreateTypeUsageNode(literal, context);
+            var typeReference = (typeNode.FirstChild as IReferenceName).Reference;
+            
+            var typeResolver = TypeResolverFactory(typeNode);
+
+            var candidates = typeReference.GetAllNames()
+                .SelectMany(typeName => typeResolver(typeName))
+                .OfType<ITypeElement>()
+                .Where(typeElement => ImportTypeUtil.TypeIsVisible(typeElement, typeNode)).ToArray();
```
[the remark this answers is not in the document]
krasnotsvetov: Nested classes don't work via string overload method, but work via generic function.
Generic classes don't work at all.
In this case, GetComponent() return null too.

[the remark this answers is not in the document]
TessenR: How does visibility check work for regular classes from other assemblies?
Project1.File1.cs
```cs
internal class Foo { }
```
Project2.TestClass.cs
```cs
class Test : MonoBehavior
{
  public object M() => GetComponent("Foo");
}
```
(no internals visible to between the projects)

resharper/src/resharper-unity/CSharp/Daemon/Errors/CSharpErrors.xml
```diff
@@ -180,6 +204,53 @@
     <Range>EqualityExpression.GetHighlightingRange()</Range>
     <Behavour overlapResolvePolicy="WARNING" />
   </Warning>
+  
+  <Warning name="UseExplicitTypeInsteadOfStringUsing" configurableSeverity="Unity.UseExplicitTypeInsteadOfStringUsing">
```
citizenmatt: `PreferGenericMethodOverload`

resharper/src/resharper-unity/CSharp/Daemon/Errors/CSharpErrors.xml
```diff
@@ -180,6 +204,53 @@
...
+    <Parameter type="IInvocationExpression" name="invocationMethod" />
+    <Parameter type="string" name="methodName" isValid="skip"/>
+    <Parameter type="string" name="stringLiteral" isValid="skip"/>
+    <Parameter type="IExpression" name="argument" />
+    <Parameter type="ITypeElement[]" name="availableTypes" isValid="skip"/>
+    <Message value="Using the string for the component type variants results in additional overhead and is not safe. 
+    Suggestion is transform from '.{0}(\&quot;{1}\&quot;)' to '.{0}&lt;{1}&gt;()'.">
```
citizenmatt: "Prefer generic method overload". If you really want to add the method: "Prefer generic method overload 'GetComponent<Foo>()'"

resharper/src/resharper-unity/CSharp/Daemon/Errors/CSharpErrors.xml
```diff
@@ -52,6 +52,30 @@
...
+        <Title>Correct type is not found</Title>
+        <Description>The type could not be resolved from string</Description>
+      </Tag>
+
+      <Tag externalName="AmbiguousUnityMonoBehaviourDefinitionWarning.HIGHLIGHTING_ID"
+           default="WARNING">
+        <Title>Inheritor of 'MonoBehaviour' with same name defined twice in different namespaces</Title>
+        <Description>Its dangerous to use name of this type in methods like 'GetComponent' due to unexpected behaviour</Description>
+      </Tag>
+
```
citizenmatt: You can probably use the `MultipleResolveCandidatesInTextWarning` here. It'll say:

> Ambiguous reference: 
> Thing.Foo
> Whatever.Foo
> match

Alternatively:

Title: "Ambiguous MonoBehaviour reference: {0} match"
Description: "The string literal overload of methods such as GameObject.AddComponent, GameObject.GetComponent and ScriptableObject.CreateInstance can use non-qualified names, and cause ambiguous references, where one class is selected at random at runtime. Prefer fully qualified names, or the generic method overload"

Also, does this apply to `ScriptableObject.CreateInstance`?

resharper/src/resharper-unity/CSharp/Daemon/Errors/CSharpErrors.xml
IvanPashchenko: I agree on highlighting just the string literal.

resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/PreferGenericMethodOverloadAnalyzer.cs
```diff
@@ -0,0 +1,191 @@
...
+        public PreferGenericMethodOverloadAnalyzer(UnityApi unityApi)
+            : base(unityApi)
+        {
+        }
+
+        protected override void Analyze(IInvocationExpression expression, ElementProblemAnalyzerData data,
+            IHighlightingConsumer consumer)
+        {
+            if (expression.RPar == null) return;
+            if (expression.ContainsPreprocessorDirectives()) return;
```
TessenR: 1. This check is considerably more expensive than the subsequent ones that are likely to filter out 95%+ of invocations that shouldn't be processed. Consider moving it somewhere later in the method.

2. It probably should only be checked for `PreferGenericMethodOverloadWarning`. Other warnings e.g. `UnknownTypeInStringLiteralWarning` look valid regardless of presence of inactive preprocessor branches. This would also solve the first problem.

PR #766, opened 2018-09-04
EditorPlugin pre-5.6 is build with Unity libs 4.7 and define UNITY_4_7

van800:
continuation of discussion in https://github.com/JetBrains/resharper-unity/pull/762


citizenmatt:
Cool, looks good. Should we also compile the (released) full version against 5.6 assemblies?

van800:
We can, but since now define is UNITY_2017_3 and referenced libs are from 2017.3.0f3, so it looks good enough to me.

PR #768, opened 2018-09-05
Add custom naming style for serialised fields

citizenmatt:
This PR will recognise Unity serialised fields as an entity that has its own naming style, rather than using the existing public or private field naming styles. By default, the naming style is `lowerCamelCase`, and if a serialised field doesn't match, it will receive a highlight and a quick fix to rename.

The naming style can be changed (`UpperCamelCase`, `lowerCamlCase_UnderscoreTolerant`, etc.), however, this is not available in the standard C# Naming Code Style tab (see [RIDER-8339](https://youtrack.jetbrains.com/issues/RIDER-8339)), but is available as a drop down in the Unity settings tab.

### DIFF
--- a/resharper/src/resharper-unity/CSharp/Psi/Naming/Elements/UnityNamedElement.cs
+++ b/resharper/src/resharper-unity/CSharp/Psi/Naming/Elements/UnityNamedElement.cs
@@ -1,0 +1,48 @@
+using System;
+using JetBrains.Annotations;
+using JetBrains.ProjectModel;
+using JetBrains.ReSharper.Psi;
+using JetBrains.ReSharper.Psi.CSharp;
+using JetBrains.ReSharper.Psi.Naming.Elements;
+using JetBrains.ReSharper.Psi.Naming.Settings;
+
+namespace JetBrains.ReSharper.Plugins.Unity.CSharp.Psi.Naming.Elements
+{
+    // We could pass null here, but we only actually support C#...
+    [NamedElementsBag(typeof(CSharpLanguage))]
+    public class UnityNamedElement : ElementKindOfElementType
+    {
+        [UsedImplicitly] public static readonly IElementKind SERIALISED_FIELD =
+            new UnityNamedElement("UNITY_SERIALISED_FIELD", "Unity serialized field", IsSerialisedField, new NamingRule
+            {
+                NamingStyleKind = NamingStyleKinds.aaBb
+            });
+
+        private readonly NamingRule myNamingRule;
+
+        protected UnityNamedElement(string name, string presentableName, Func<IDeclaredElement, bool> isApplicable,
+                        NamingRule namingRule)
+            : base(name, presentableName, isApplicable)
+        {
+            myNamingRule = namingRule;
+        }
+
+        public override PsiLanguageType Language => CSharpLanguage.Instance;
+
+        // This doesn't really do anything useful. See UnityNamingRuleDefaultSettings
+        public override NamingRule GetDefaultRule() => myNamingRule;
+
+        private static bool IsSerialisedField(IDeclaredElement declaredElement)
+        {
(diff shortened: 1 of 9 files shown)


englerj:
In my Unity projects, I use UpperCamelCase for public fields and _lowerCamelCase for private [serialized] fields. After updating to Rider 2018.3 EAP, this change is causing half of my serialized fields to (correctly) display "Inconsistent Naming" warnings.

I would prefer to keep my current naming style, but there doesn't appear to be a way to disable this warning. Since it is not visible in the C# Naming Code Style tab, I can't remove this style alone, but since it still acts as part of the C# Naming Code Style inspection, the only way to disable the warning would be to disable the _Inconsistent Naming_ inspection altogether (which i'd prefer to keep active). Without the RIDER-8339 enhancement, i feel it necessary that there be a way to disable this check (e.g. use standard Naming Code Style).

p.s. Wasn't sure if I needed to raise a separate issue for this or not. Please advise where i should move this if this is not the correct play.

SugoiDev:
I think all we need for now is for this rule to respect the **Prefix** set in the xml. So, something like this would be respected:

`
<s:String x:Key="(...) Prefix="_" Suffix="" Style="aaBb" (...)</s:String>
`

I know some prefer to use **m_** as prefix, so respecting the prefix is probably very relevant.

It would most likely be a good idea to also respect the **AccessRightKinds** as well, as it would also fix #859

SugoiDev:
Well, what do you know. I just tested today and it seems the settings are respected when editing the dotsettings directly.

```

	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/UserRules/=5f0fdb63_002Dc892_002D4d2c_002D9324_002D15c80b22a7ef/@EntryIndexedValue">&lt;Policy&gt;&lt;Descriptor Staticness="Instance" AccessRightKinds="Public" Description="Unity serialized field"&gt;&lt;ElementKinds&gt;&lt;Kind Name="UNITY_SERIALISED_FIELD" /&gt;&lt;/ElementKinds&gt;&lt;/Descriptor&gt;&lt;Policy Inspect="True" Prefix="" Suffix="" Style="AaBb" /&gt;&lt;/Policy&gt;</s:String>
	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/UserRules/=5f0fdb63_002Dc892_002D4d2c_002D9324_002D15c80b22a7ef/@EntryIndexedValue">&lt;Policy&gt;&lt;Descriptor Staticness="Instance" AccessRightKinds="Private" Description="Unity serialized field"&gt;&lt;ElementKinds&gt;&lt;Kind Name="UNITY_SERIALISED_FIELD" /&gt;&lt;/ElementKinds&gt;&lt;/Descriptor&gt;&lt;Policy Inspect="True" Prefix="_" Suffix="" Style="aaBb" /&gt;&lt;/Policy&gt;</s:String>

```

With the settings above, I managed to get my public fields to be **AaBb** and my private ones to be **_aaBb**, so this might be a temporary solution to #859 until the UI is updated.

citizenmatt:
Just as a follow up; while this works, it's not quite right - you shouldn't use the same guid for each rule.

Also, the UI for editing the serialised field rule will include prefix/suffix and a checkbox to disable inspections in the final release of 2018.3. It doesn't allow matching public/private. I'm planning on addressing RIDER-8339 for 2019.1.

jjenn:
@citizenmatt  If I want to change `Naming style for serialized fields` in `ReSharper` on `Visual Studio 2019`, what should I do? There is no `Languages & Frameworks` tab to choose.

citizenmatt:
In ReSharper, you can use the standard [C# naming style options page](https://www.jetbrains.com/help/resharper/Coding_Assistance__Naming_Style.html). Here you can add custom rules that include Unity serialised fields.

PR #772, opened 2018-09-06
Added inspection for incorrect method bodies of DrawGizmo decorated methods

JurjenBiewenga:


resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/DrawGizmoAttributeProblemAnalyzer.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        private void CheckMethodDeclaration(IMethodDeclaration methodDeclaration, IAttribute element, IHighlightingConsumer consumer)
+        {
+            if (methodDeclaration == null)
+                return;
+
+            var predefinedType = myPredefinedTypeCache.GetOrCreatePredefinedType(element.GetPsiModule());
+            var gizmoTypeName = TypeFactory.CreateTypeByCLRName("UnityEditor.GizmoType", predefinedType.Module);
+            var componentName = TypeFactory.CreateTypeByCLRName("UnityEngine.Component", predefinedType.Module);
```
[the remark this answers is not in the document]
JurjenBiewenga: I need a `IDeclaredType` while KnownTypes gives me an `ICLRTypeName` which is not castable.

resharper/test/data/CSharp/Daemon/Stages/Analysis/DrawGizmoAttributeProblemAnalyzer.cs
```diff
@@ -0,0 +1,30 @@
...
+	
+    [DrawGizmo]
+    public static void WrongTypeParameters<T>(Component arg1, GizmoType arg2) { }
+
+    [DrawGizmo]
+    public static void JustRight(Component arg1, GizmoType arg2) { }
+
+    [DrawGizmo]
+    private static void JustRight2(MonoBehaviour arg1, GizmoType arg2) { }
+}
```
citizenmatt: Can you add a couple more tests where the first parameter are derived types, please? Just add a couple of dummy classes into this file and derive from `MonoBehaviour` and `Component`.

resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/DrawGizmoAttributeProblemAnalyzer.cs
```diff
@@ -47,49 +47,43 @@ private void CheckMethodDeclaration(IMethodDeclaration methodDeclaration, IAttri
...
-            else
-            {
-                var expectedDeclaration = new MethodSignature(predefinedType.Void, true,
-                    new[] {componentName, gizmoTypeName},
-                    new[] {"scr", "gizmoType"});
-                consumer.AddHighlighting(new InvalidParametersWarning(methodDeclaration, expectedDeclaration));
+                Log.Root.Log(LoggingLevel.INFO, string.Format("component {0}:{1}:{2}", parameters[0].Type.GetTypeElement(), componentType.GetTypeElement(), parameters[0].Type.GetTypeElement()
+                    ?.IsDescendantOf(componentType.GetTypeElement())));
+                Log.Root.Log(LoggingLevel.INFO, string.Format("gizmo {0}:{1}:{2}", parameters[1].Type.GetTypeElement(), gizmoType.GetTypeElement(), parameters[1].Type.GetTypeElement()?.Equals(gizmoType?.GetTypeElement())));
+
```
citizenmatt: Let's lose the logs. We don't need them for release.

And for the future, you can inject `ILogger` into the constructor of components such as this and the container will provide you the correct instance, so you wouldn't use `Log.Root` directly.

resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/DrawGizmoAttributeProblemAnalyzer.cs
```diff
@@ -47,49 +47,43 @@ private void CheckMethodDeclaration(IMethodDeclaration methodDeclaration, IAttri
...
             if (methodDeclaration == null)
                 return;
-
+            
             var predefinedType = myPredefinedTypeCache.GetOrCreatePredefinedType(element.GetPsiModule());
-            var gizmoTypeName = TypeFactory.CreateTypeByCLRName("UnityEditor.GizmoType", predefinedType.Module);
-            var componentName = TypeFactory.CreateTypeByCLRName("UnityEngine.Component", predefinedType.Module);
+            var gizmoType = TypeFactory.CreateTypeByCLRName("UnityEditor.GizmoType", predefinedType.Module);
+            var componentType = TypeFactory.CreateTypeByCLRName("UnityEngine.Component", predefinedType.Module);
+
```
citizenmatt: The last comment is outdated, but what I meant was:

```
TypeFactory.CreateTypeByCLRName(KnownTypes.Component, ...)
```

You'll have to add `KnownTypes.GizmoType`. Using this string overload will create an instance of `ClrTypeName` with what you pass in. Using `KnownTypes` just shares those instances.

resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/DrawGizmoAttributeProblemAnalyzer.cs
```diff
@@ -0,0 +1,89 @@
...
+                    Log.Root.Log(LoggingLevel.INFO, "First matched");
+                    if (parameters[1].Type.GetTypeElement()?.Equals(gizmoType?.GetTypeElement()) == true)
+                    {
+                        Log.Root.Log(LoggingLevel.INFO, "second matched");
+                        match &= ~MethodSignatureMatch.IncorrectParameters;
+                    }
+                
+                    expectedDeclaration = new MethodSignature(predefinedType.Void, true,
+                        new[] {parameters[0].Type, gizmoType},
+                        new[] {parameters[0].DeclaredName, parameters[1].DeclaredName});
```
citizenmatt: I really like this. Means that the quick fix if the parameters are wrong will have the correct type for the first parameter. But it also means that I want it for more scenarios, too :)

Firstly, if there are one or more parameters, find the first one that derives from `Component` or `MonoBehaviour` and use this type to update the `MethodSignature`. So if someone has `Foo(MyComponent component, int whatever, string yeah, float really, double wrong)`, the inspections and quick fix will show "Change parameters to '(MyComponent, GizmoType)'" instead of losing information with "Change to '(Component, GizmoType)'". And if they do `Foo(GizmoType, MyComponent)`, or `Foo(MyComponent, int)' we'll show them the correct type for `Component` too.

Secondly, check that there are two parameters, and that they are the correct types (probably worth setting a flag from the first check for the component type).

And add test cases:
* Too many parameters with the component type in the first position
* Too many parameters with the component type NOT in the first position
* Correct first parameter and incorrect second parameter

resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/DrawGizmoAttributeProblemAnalyzer.cs
```diff
@@ -0,0 +1,83 @@
...
+                new[] {"component", "gizmoType"});
+
+            var match = expectedDeclaration.Match(methodDeclaration);
+            
+            if (methodDeclaration.Params.ParameterDeclarations.Count == 2)
+            {
+                var parameters = methodDeclaration.Params.ParameterDeclarations;
+
+                if (parameters[0].Type.GetTypeElement()
+                        ?.IsDescendantOf(componentType.GetTypeElement()) == true)
```
citizenmatt: Looks like `IsDescendantOf` returns `true` if the derived type is equal to the base type. I don't think we want that - can you add a DrawGizmo for `MonoBehaviour` or `Component` directly?

JurjenBiewenga: Unity doesn't appear to support gizmos on the base type of your class however if we add in that it is incorrect you'll kind of get stuck in an infinite cycle of it being wrong until you figure out that it can't be the base type. So we'll have to add an extra inspection for that I think.

resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/DrawGizmoAttributeProblemAnalyzer.cs
```diff
@@ -53,27 +53,51 @@ private void CheckMethodDeclaration(IMethodDeclaration methodDeclaration, IAttri
...
             if (methodDeclaration.Params.ParameterDeclarations.Count == 2)
             {
                 var parameters = methodDeclaration.Params.ParameterDeclarations;
 
-                if (parameters[0].Type.GetTypeElement()
-                        ?.IsDescendantOf(componentType.GetTypeElement()) == true)
+                if (firstParamCorrect)
                 {
-                    if (parameters[1].Type.GetTypeElement()?.Equals(gizmoType?.GetTypeElement()) == true)
+                    if (secondParamCorrect)
```
citizenmatt: Minor nitpick - merge the `if` conditions

citizenmatt:
OK, next step is the extra inspection if just the type of the first parameter is incorrect. This is fairly straightforward to add - you've already got the logic, we just need to create a new warning type. This involves adding two sections into the [`CSharpErrors.xml`](https://github.com/JetBrains/resharper-unity/blob/183/resharper/src/resharper-unity/CSharp/Daemon/Errors/CSharpErrors.xml) file. There are some comments in there that will help.

1. Add a new `Tag` item to `SeverityConfiguration`. This is the id, title + description for configurable severities, and is the data for the inspection severity dialog. Add a short title, provide a slightly longer description and give it a default severity (most of ours seem to be `WARNING`, which is fine). You also need to give it an `externalName`. By convention, we use `{inspection_name}Warning.HIGLIGHTING_ID`, where `inspection_name` is defined below.
2. Add a new `Warning` element, with a name to match the `inspection_name` you used above. Set the `configurableSeverity` to be `Unity.{inspection_name}`. This node is used to generate the warning class that you create in the analyser and pass to the highlighting consumer. The rest of the XML tags are documented in comments, but it's still a bit tricky.

Add `Parameter` elements to pass parameters to the constructor of your warning class. This is likely to be something like `IParameter` - the tree node that you're going to apply the inspection to.

The `Message` element is what is shown in the inspection tooltip. It can be a string literal, or can contain `{0}` and `{1}` elements, with the values coming from nested `Argument` elements. See the `IncorrectSignature` warning.

`Range` is the range of the tree node that gets highlighted. You'd likely do something like `Parameter.GetHighlightingRange()`.

And finally, add `Behavour` (note the spelling!) with `overlapPolicy="WARNING"`. This just defines the overlap policy which is what happens if two inspections are on the same range - who wins.

Now build the project and it will generate a new warning class for you, as part of the `CSharp.generated.cs` file. If there are errors, you can open the file, see what you need to fix and figure out what the xml file should have to generate the correct code.

It feels like a lot, especially when I've just rewritten the contents of the comments here, but it's really quite straightforward, especially after you've done it once.

resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/DrawGizmoAttributeProblemAnalyzer.cs
```diff
@@ -90,18 +88,18 @@ private void CheckMethodDeclaration(IMethodDeclaration methodDeclaration, IAttri
...
             if (methodDeclaration.Params.ParameterDeclarations.Count == 2)
             {
-                var parameters = methodDeclaration.Params.ParameterDeclarations;
-
-                if (firstParamCorrect)
+                if (firstParamCorrect && secondParamCorrect)
+                {
+                    match &= ~MethodSignatureMatch.IncorrectParameters;
+                }
+                else if (!firstParamCorrect && match == MethodSignatureMatch.IncorrectParameters)
```
citizenmatt: I think this also needs a check for `secondParamCorrect` being true.

citizenmatt:
Hooray! Congrats on adding on a cool new feature :tada:

Issue #773, opened 2018-09-07
Stop running tests in Unity by Stopping them in Rider


van800:
Example of long running test
```
        [UnityTest]
        public IEnumerator EditModeTest1WithEnueratorPasses()
        {
            for (int i = 0; i < 10000; i++)
            {
                yield return null;
            }
        }
```

PR #774, opened 2018-09-07
[WIP] Allow Abort unit tests

van800:
https://github.com/JetBrains/resharper-unity/issues/773


van800:
I am not sure about lifetime, which is passed to UnityEditorTestLauncher, could you check @citizenmatt please?

citizenmatt:
Yeah, using `connectionLifetime` will effectively leak memory - `AdviseNotNull` will add an unsubscribe handler to `connectionLifetime` for each new launch, so every time we run tests, a handler gets added, but no-one is removing it, so it will keep a reference to the `Abort` property alive.

We need a lifetime that lasts for the duration of the test run, so I think `UnityEditorTestLauncher` should create a lifetime definition with `connectionLifetime` as the parent and terminate it in `RunFinished`. So when the run finishes, the terminated lifetime will remove the subscription from `Abort`.

unity/EditorPlugin/UnitTesting/UnityEditorTestLauncher.cs
```diff
@@ -10,21 +12,24 @@
...
     private const string MRunfinishedevent = "m_RunFinishedEvent";
     private const string MTestfinishedevent = "m_TestFinishedEvent";
     private const string MEditmoderunner = "m_EditModeRunner";
+    private const string MRunner = "m_Runner";
     private const string TestNames = "testNames";
     private const string TestNames56 = "names";
     private const string EditModeLauncher = "UnityEditor.TestTools.TestRunner.EditModeLauncher";
+    private const string PlayModeLauncher = "UnityEditor.TestTools.TestRunner.PlaymodeLauncher";
     private const string TestRunnerFilter = "UnityEngine.TestTools.TestRunner.GUI.TestRunnerFilter";
     private const string TestPlatform = "UnityEngine.TestTools.TestPlatform";
```
citizenmatt: I'd consider inlining these constants if they're only used in one place. Otherwise, you only have to look them up when reading the code.

unity/EditorPlugin/UnitTesting/UnityEditorTestLauncher.cs
```diff
@@ -58,9 +64,10 @@ public void TryLaunchUnitTests()
         string testEditorAssemblyProperties =  testEditorAssembly.GetType().GetProperties().Select(a=>a.Name).Aggregate((a, b)=>a+ ", "+b);
         string testEngineAssemblyProperties = testEngineAssembly.GetType().GetProperties().Select(a=>a.Name).Aggregate((a, b)=>a+ ", "+b);
```
citizenmatt: I think this just gets a list of the properties on `Assembly` - is there a missing type name in the call to `GetType()`?

unity/EditorPlugin/UnitTesting/UnityEditorTestLauncher.cs
```diff
@@ -155,6 +163,30 @@ public void TryLaunchUnitTests()
...
+        myLaunch.Abort.AdviseNotNull(myLifetime, _ =>
+        {
+          ourLogger.Verbose($"Call {StopRun} method via reflection.");
+          try
+          {
+            stopRunMethod.Invoke(unityTestAssemblyRunner, null);
+          }
+          catch (Exception)
+          {
+            ourLogger.Verbose($"Call {StopRun} method failed.");
```
citizenmatt: Log the exception, too?

unity/EditorPlugin/UnitTesting/UnityEditorTestLauncher.cs
```diff
@@ -58,9 +64,10 @@ public void TryLaunchUnitTests()
         string testEditorAssemblyProperties =  testEditorAssembly.GetType().GetProperties().Select(a=>a.Name).Aggregate((a, b)=>a+ ", "+b);
         string testEngineAssemblyProperties = testEngineAssembly.GetType().GetProperties().Select(a=>a.Name).Aggregate((a, b)=>a+ ", "+b);
 
-        var launcherType = testEditorAssembly.GetType(EditModeLauncher);
+        var launcherTypeString = myLaunch.TestMode == TestMode.Edit ? EditModeLauncher : PlayModeLauncher;
+        var launcherType = testEditorAssembly.GetType(launcherTypeString);
         if (launcherType == null)
-          throw new NullReferenceException($"Could not find {EditModeLauncher} among {testEditorAssemblyProperties}");
+          throw new NullReferenceException($"Could not find {launcherTypeString} among {testEditorAssemblyProperties}");
```
citizenmatt: Perhaps use `Assertion.AssertNotNull` here instead of throwing `NullReferenceException`? 

That would change behaviour, though - `AssertNotNull` doesn't throw in release mode. If we expect this to never be null, that's ok, but if we expect it to be null, perhaps we should return gracefully instead of throwing?

PR #778, opened 2018-09-12
Quick fix for NonAlloc API

krasnotsvetov:
Functions like `SphereCast`, `RaycastAll`, `CapsuleCast` e.t.c has NonAlloc version, analyzer generates warnings when user uses this functions. In some cases, where we can improve user code without unexpected affect quickfix is provided.


citizenmatt:
NOTE: This PR was reverted

PR #783, opened 2018-09-14
Quick fix for access properties by id instead of string

krasnotsvetov:
Access properties in `Shader`, `Material`, `Animator` by `int` is faster than using `string` overload.

 The property IDs created from string hashes are deterministic over the course of a single run, so this quick fix creates a private static readonly field in class and reuse this field for overload.
Example:
```
public class SimpleTest
{
    public void Method(Material material)
    {
        material.SetFloat("test", 10.0f);
    }
}
```

is transformed to:

```
public class SimpleTest
{
    private static readonly int Test = Shader.PropertyToID("test");
    public void Method(Material material)
    {
        material.SetFloat(Test, 10.0f);
    }
}
```

### DIFF
--- a/resharper/src/resharper-unity/CSharp/Feature/Services/QuickFixes/PreferAddressByIdToGraphicsParamsQuickFix.cs
+++ b/resharper/src/resharper-unity/CSharp/Feature/Services/QuickFixes/PreferAddressByIdToGraphicsParamsQuickFix.cs
@@ -1,0 +1,148 @@
+using System;
+using JetBrains.Application.Progress;
+using JetBrains.ProjectModel;
+using JetBrains.ReSharper.Feature.Services.QuickFixes;
+using JetBrains.ReSharper.Plugins.Unity.CSharp.Daemon.Errors;
+using JetBrains.ReSharper.Psi;
+using JetBrains.ReSharper.Psi.CSharp;
+using JetBrains.ReSharper.Psi.CSharp.Tree;
+using JetBrains.ReSharper.Psi.Naming.Extentions;
+using JetBrains.ReSharper.Psi.Naming.Impl;
+using JetBrains.ReSharper.Psi.Naming.Settings;
+using JetBrains.ReSharper.Psi.Tree;
+using JetBrains.TextControl;
+using JetBrains.Util;
+
+namespace JetBrains.ReSharper.Plugins.Unity.CSharp.Feature.Services.QuickFixes
+{
+    [QuickFix]
+    public class PreferAddressByIdToGraphicsParamsQuickFix : QuickFixBase
+    {
+        private IInvocationExpression myInvocationExpression;
+        private ICSharpArgument myArgument;
+        private string myLiteral;
+        private string myMapFuntion;
+        private string myTypeName; 
+        
+        public PreferAddressByIdToGraphicsParamsQuickFix(PreferAddressByIdToGraphicsParamsWarning warning)
+        {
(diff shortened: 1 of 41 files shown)


krasnotsvetov:
Add test with undescore name

citizenmatt:
LGTM!

resharper/test/data/CSharp/Intentions/QuickFixes/PreferAddressByIdToGraphicsParams/InvalidLiteralForPropertyNameTest.cs.gold
```diff
@@ -0,0 +1,14 @@
+﻿using UnityEngine;
+
+namespace DefaultNamespace
+{
+    public class InvalidLiteralForProperyNameTest
+    {
+      private static readonly int Property = Shader.PropertyToID("Property");
```
TessenR:
```cs
material.SetFloat("f{caret}loat", 10.0f);`
```
is converted to 

```cs
private static readonly int Property = Shader.PropertyToID("Property");
...
   material.SetFloat({caret}Property, 10.0f);
```

Are these code samples really equivalent?

resharper/src/resharper-unity/CSharp/Feature/Services/QuickFixes/PreferAddressByIdToGraphicsParamsQuickFix.cs
```diff
@@ -0,0 +1,258 @@
...
+            var argument = factory.CreateArgument(ParameterKind.VALUE, myArgument.NameIdentifier?.Name, referenceExpression);
+            myArgument.ReplaceBy(argument);
+            return null;
+        } 
+
+        private static IDeclaredElement TryFindIdDeclarationInClassLikeDeclaration(IClassLikeDeclaration declaration, string propertyName,
+            string requiredTypeName, string requiredMethodName, out string name)
+        {
+            name = null;
+            foreach (var member in declaration.ClassMemberDeclarations)
```
TessenR: This code only checks declarations syntaxically located in the passed declaration.
I belive that my original example still won't reuse existing property, could you please add a test for it?
```cs
public partial class TestClass
{
  public void Method(Material material)
  {
    material.SetFloat("te{caret}st", 10.0f);
  }
}

public partial class TestClass
{
  private static readonly int Test = Shader.PropertyToID("test");
}
```

You can check all declarations of a type like this
```cs
foreach (var declarationPart in declaration.DeclaredElement.GetDeclarations())
```

resharper/src/resharper-unity/CSharp/Feature/Services/QuickFixes/PreferAddressByIdToGraphicsParamsQuickFix.cs
```diff
@@ -0,0 +1,258 @@
...
+            myMapFunction = warning.MapFunction;
+            myTypeName = warning.TypeName;
+        }
+        
+        protected override Action<ITextControl> ExecutePsiTransaction(ISolution solution, IProgressIndicator progress)
+        {
+            var factory = CSharpElementFactory.GetInstance(myInvocationExpression);
+
+            //  try find declaration where string to id conversation is done. If we don't find it, create in top-level class
+            var idDeclaration = TryFindDeclaration(myInvocationExpression, myGraphicsPropertyName, myTypeName, myMapFunction, out var name);
```
TessenR: Looks like `out var name` is always overwritten before usage. Can we get rid of it?

PR #784, opened 2018-09-14
Quick fix for NonAlloc API

krasnotsvetov:
Functions like SphereCast, RaycastAll, CapsuleCast e.t.c has NonAlloc version, analyzer generates warnings when user uses this functions. In some cases, where we can improve user code without unexpected affect quickfix is provided.


resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/PreferNonAllocApiAnalyzer.cs
```diff
@@ -0,0 +1,109 @@
...
+            
+            string newName; // xxx[All] -> xxxNonAlloc
+            if (originName.EndsWith("All"))
+            {
+                newName = originName.Substring(0, originName.Length - 3) + "NonAlloc";
+            }
+            else
+            {
+                newName = originName + "NonAlloc";
+            }
```
citizenmatt: Move this section below the check for type, or we'll be doing this for every method invocation in the project.

resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/PreferNonAllocApiAnalyzer.cs
```diff
@@ -0,0 +1,109 @@
...
+            // drop out all other invocation
+            if (!containingType.GetClrName().Equals(KnownTypes.Physics) &&
+                !containingType.GetClrName().Equals(KnownTypes.Physics2D))
+                return null;
+
+            var type = TypeFactory.CreateType(containingType);
+            var table = type.GetSymbolTable(expression.PsiModule).Filter(
+                new AccessRightsFilter(new DefaultAccessContext(expression)),
+                new ExactNameFilter(newName),
+                new PredicateFilter(t => MatchSignatureAllocToNonAlloc(method, t.GetDeclaredElement() as IMethod)));
```
citizenmatt: Is this heuristic good enough to capture all of the methods we need? Can we rely on naming and parameter size +1? Would it be easier to just have a map of allocating to non-allocating methods?

PR #793, opened 2018-09-18
Minor Unity Explorer fixes

citizenmatt:
* Show Unity Explorer (and auto-save notification) for non-generated files
* Fixes Unity Explorer not showing up for some projects in Linux


citizenmatt:
@van800 can you just check the tweaks I've made to `isUnityProject` please?

I've got rid of the check for the `Library` folder, so we can now use just `isUnityProject` or `isUnityProjectFolder`. No-one's using `isUnitySidecarProject`, although I've left it in for now.

The "switch UI mode" functionality is currently using `isUnityGeneratedProject`. Should this be changed to `isUnityProject`?

PR #797, opened 2018-09-19
WIP: Avoid multiple unnecessary property accesses inspection,  instantiation of GameObject inspection

krasnotsvetov:
# Avoid multiple unnecessary property accesses inspection
Accessing properties of unity's built-in component results in native call. This is performance overhead.

Consider an example:
```
public class Behavior : MonoBehaviour
{
  void Foo(Transform transform)
  {
    // Each position access will require getting the transform then calling the getter for the position property. 
    transform.position = transform.position + new Vector3(0.0f, 1.0f, 0.0f);
    transform.position = transform.position / 2;
    //etc. etc. some transform.position manipulation
  }
}
```
Value of property can be cached. 
```
public class Behavior : MonoBehaviour
{
  void Foo(Transform transform)
  {    
    var position = transform.position; 
    position = position + new Vector3(0.0f, 1.0f, 0.0f);
    position = position / 2;
    transform.position = position;
  }
}
```

This inspection generates warnings for cases of multiply property accessing and provide a quick fix.

# Instantiation of GameObject inspection

Using method overload with parent parameter avoids creating a new transform hierarchy and then moving it and instead quickly inserts it into the existing transform hierarchy

```
GameObject go = (GameObject) GameObject.Instantiate(exampleGO);
go.transform.par{caret}ent = root.transform;
```
transforms to
```
GameObject go = (GameObject) GameObject.Instantiate(exampleGO, root.transform, true);
```


resharper/src/resharper-unity/CSharp/Daemon/Errors/CSharpErrors.xml
```diff
@@ -46,6 +46,13 @@
         <Description>Parameter should be derived from `Component`.</Description>
       </Tag>
 	  
+
+      <Tag externalName="InefficientPropertyAccessWarning.HIGHLIGHTING_ID"
+           default="WARNING">
+        <Title>Access properties of built-in component is not efficient. Cache value</Title>
```
IvanPashchenko: Repeatable property access of a built-in component is inefficient

resharper/src/resharper-unity/CSharp/Daemon/Errors/CSharpErrors.xml
```diff
@@ -46,6 +46,13 @@
...
+        <Description>Access properties of built-in component is not efficient due to calling native methods.</Description>
```
IvanPashchenko: Built-in component property access results in native method calls and should be cached

resharper/src/resharper-unity/CSharp/Daemon/Errors/CSharpErrors.xml
```diff
@@ -245,6 +252,15 @@
     <Range>Argument.GetHighlightingRange()</Range>
     <Behavour overlapResolvePolicy="NONE" />
   </Warning>
+
+  <Warning name="InefficientPropertyAccess" configurableSeverity="Unity.InefficientPropertyAccess">
+    <Parameter type="IReferenceExpression" name="highlightedReference"/>
+    <Parameter type="IReferenceExpression[]" name="references" isValid="skip" />
+    <Parameter type="bool" name="canApplyFix" />
+    <Message value="Access properties of built-in component is not efficient. Cache value"/>
```
IvanPashchenko: Same as upper description

resharper/src/resharper-unity/KnownTypes.cs
```diff
@@ -21,6 +21,7 @@ public static class KnownTypes
         public static readonly IClrTypeName RuntimeInitializeOnLoadMethodAttribute = new ClrTypeName("UnityEngine.RuntimeInitializeOnLoadMethodAttribute");
         public static readonly IClrTypeName ScriptableObject = new ClrTypeName("UnityEngine.ScriptableObject");
         public static readonly IClrTypeName SerializeField = new ClrTypeName("UnityEngine.SerializeField");
+        public static readonly IClrTypeName Transform = new ClrTypeName("UnityEngine.Transform");
```
IvanPashchenko: remove if unused

resharper/src/resharper-unity/Utils/UnityComponentRelatedExpressionFinder.cs
```diff
@@ -0,0 +1,372 @@
+using System.Collections.Generic;
+using JetBrains.Annotations;
+using JetBrains.ReSharper.Psi;
+using JetBrains.ReSharper.Psi.CSharp.Tree;
+using JetBrains.ReSharper.Psi.Tree;
+using JetBrains.Util;
+using QualifierEqualityComparer = JetBrains.ReSharper.Psi.CSharp.Impl.ControlFlow.ControlFlowWeakVariableInfo.QualifierEqualityComparer;
```
TessenR: I believe that it'd be better to write specialized equality comparer instead of using an existing one from `ControlFlowWeakVariableInfo` as it wasn't supposed to be used outside of control flow inspections.

1. It does more work than needed for these inspections e.g. processing indexers in qualifiers.
2. It might change at any moment according to it's intended usages in control flow inspections with no regards of such external dependencies since they aren't supposed to be there.
3. It does not provide options for customizing behavior e.g. how to handle conditional accesses.

It looks like checking pairs of reference expressions one by one is a simple enough case to implement it separately.

Also please

resharper/test/data/CSharp/Intentions/QuickFixes/CachePropertyValue/LambdaTest.cs
```diff
@@ -0,0 +1,12 @@
+using UnityEngine;
+
+namespace DefaultNamespace
+{
+    public class LabmdaTest
+    {
+        public void Test(Transform t)
+        {
+            var lambda = () => t.position + t.position + t.posi{caret}tion;
```
TessenR: Please add tests for void-returning lambda expressions e.g.
`Action action = () => DoSomething(t.position + t.position + t.posi{caret}tion);`

and lambdas under return statements e.g.
`return DoSomething(() =>  t.position + t.position + t.posi{caret}tion);`

resharper/test/src/CSharp/Intentions/QuickFixes/CachePropertyValueQuickFixTests.cs
```diff
@@ -0,0 +1,26 @@
+using JetBrains.ReSharper.FeaturesTestFramework.Intentions;
+using JetBrains.ReSharper.Plugins.Unity.CSharp.Feature.Services.QuickFixes;
+using NUnit.Framework;
+
+namespace JetBrains.ReSharper.Plugins.Unity.Tests.CSharp.Intentions.QuickFixes
+ {
+     [TestUnity]
+     public class CachePropertyValueQuickFixTests : QuickFixTestBase<CachePropertyValueQuickFix>
```
TessenR: Please add tests with conditional accesses.
I'm especially wary of cases where conditional accesses don't match between merged expressions i.e. when `t?.position` and `t.position` are mixed.

resharper/src/resharper-unity/CSharp/Feature/Services/QuickFixes/UseInstantiateWithParentQuickFix.cs
```diff
@@ -0,0 +1,60 @@
...
+                return null;
+            
+            var factory = CSharpElementFactory.GetInstance(myInvocation);
+            
+            var first = myInvocation.Arguments[0];
+            var second = myInvocation.AddArgumentAfter(factory.CreateArgument(ParameterKind.VALUE, myNewArgument.Copy()), first);
+            var boolLiteral = factory.CreateExpression(myStayInWorldCoords.ToString().ToLower());
+            myInvocation.AddArgumentAfter(factory.CreateArgument(ParameterKind.VALUE, boolLiteral), second);
+
+            using (WriteLockCookie.Create())
```
TessenR: `ICSharpStatement.RemoveOrReplaceByEmptyStatement()`

resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/InstantiateWithoutParentProblemAnalyzer.cs
```diff
@@ -0,0 +1,169 @@
...
+            foreach (var usage in usages)
+            {
+                if (usage is IReferenceExpression referenceExpression)
+                {
+                    var fullReferenceExpression = ReferenceExpressionNavigator.GetTopByQualifierExpression(referenceExpression);
+                    if (IsUsageSetTransformParent(fullReferenceExpression, out var stayInWorldCoords, out var transform))
+                    {
+                        var finder = new TransformParentRelatedReferenceFinder(referenceExpression);
+                        var relatedExpressions = finder.GetRelatedExpressions(scope, expression).FirstOrDefault();
+                        if (relatedExpressions == null || relatedExpressions.GetTreeStartOffset() >= fullReferenceExpression.GetTreeStartOffset())
```
TessenR: I'd also recomment checking that the `SetParent` method is called in the same block. Current code seems to allow cases like this:
```cs
var variable = Instantiate();
if (something)
  variable.SetParent(firstTransform);
else
  variable.SetParent(secondTransform);
```

please add tests for such cases

resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/ComponentPropertyAccessProblemAnalyzer.cs
```diff
@@ -0,0 +1,385 @@
...
+            if (block != null)
+            {
+                if (block.Descendants<IGotoStatement>().Any())
+                    return;
+                    
+                AnalyzeStatements(block, container);
+            }
+            
+            // cache all references which is not invalidated dut to no related expression was found
+            container.InvalidateCachedValues();
```
TessenR: I'd expect that `.InvalidateCachedValues` will clear some internal caches and definitely not that it will process some references and add highlightings. If it's not possible to compe up with a good name for its current functionality it'd probably be better to just split this method in two in order to reduce confusion over what it actually does.

resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/ComponentPropertyAccessProblemAnalyzer.cs
```diff
@@ -0,0 +1,385 @@
...
+                }
+
+                myPropertiesMap.Remove(key);
+            }
+
+            private ITreeNode GetPreviousRelatedExpression(IReferenceExpression referenceExpression, ICSharpStatement readAnchor)
+            {
+                var finder = GetFinder(referenceExpression);
+                var allSequence = finder.GetRelatedExpressions(readAnchor);
+                var fromSequence = finder.GetRelatedExpressions(readAnchor, referenceExpression);
```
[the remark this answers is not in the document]
krasnotsvetov: TreeOffset order is not correct, for example:
var y = x + InvokeSmth(x + x) + x;
            ~                        ~   ~
~ can be cached only, but invocation offset is smaller and will invalidate caches.

PR #798, opened 2018-09-19
Avoid using string based method Invocation inspection

krasnotsvetov:
String based method invocation leads to harder to maintain code as well as in some cases unnecessary performance overhead


citizenmatt:
We might want to think about the severity of this inspection. It's absolutely correct, but there's no quick fix - it requires an architectural level rewrite, so a warning might be too much here. Plus, these methods are likely to be used quite extensively, so we might be overwhelming the user with un-actionable warnings.

Perhaps "hint" level?

krasnotsvetov:
What do you think about suggestion?

citizenmatt:
It's still quite intrusive, and if there are a lot of these usages, we'll be overwhelming the user again with unactionable issues. We're basically warning them about an architectural issue, which isn't easily fixed. I don't think we should nag them here. Hint is our least intrusive inspection.

krasnotsvetov:
Ok, let's move it hints. When we decided how to design API for generated class, we can seperate Invoke and InvokeRepeatable from Send* and Broadcast and provide quickfix with warning.

PR #803, opened 2018-09-21
Add inspection for instantiation of GameObject

IvanPashchenko:


resharper/src/resharper-unity/CSharp/Daemon/Errors/CSharpErrors.xml
```diff
@@ -64,6 +64,12 @@
         <Description>Using the string for the component type variants results in additional overhead that should not be incurred.</Description>
       </Tag>
 
+      <Tag externalName="InstantiateWithoutParentWarning.HIGHLIGHTING_ID"
+           default="WARNING">
+        <Title>Use overload with transform parameter instead of set transform parent later</Title>
```
IvanPashchenko: Let's move highlighter to the problematic invocation itself (SetParent) and change the text to reflect that the issue is really in setting the parent later.

The proper title then would be "Separating the initialization and setting the parent object is ineffecient" or something like that. cc @citizenmatt WDYT?

citizenmatt: Agreed on location of warning.

> Inefficient call to 'SetParent'. Combine with object creation

And explain why in the description (shown in the inspection severity dialog) and a code wiki page?

krasnotsvetov:
Closed, because this branch is merged with repeated property access inspection

PR #805, opened 2018-09-24
[WIP] Various build updates

citizenmatt:
Build improvements for .csproj files:

* Fix warnings about mismatched assembly versions
* Fix some unused variable warnings
* Clean up .csproj files - proper imports for .net sdk, and remove Unix only exclusions

And for gradle files:

* Improve skipping dotnet tasks
* Only validate plugin.xml if it's changed
* Split CI and model into own files
* Set logging level for msbuild + nuget restore based on gradle logging level

Not yet done. For discussion:

I'd like to move the unzipped SDK folders to the gradle cache, so they work like IJ does with its SDK. This would reduce the amount of _stuff_ in the `build` folders, which helps if you have multiple checkouts.

- [ ] Unzip R# SDK into gradle cache folder, rather than build folder (like IJ does with it's SDK)
- [ ] Unzip Rider SDK into gradle cache folder, too. This would change how rdgen works. The task would have to be added in an "after evaluate" step, as we don't know the Rider SDK's gradle cache location until "after evaluate"
- [ ] Rewrite `backend.gradle` as `dotnet.gradle` and change the dependencies for `runIde` to only build Rider. `buildPlugin` would continue to build everything. This will clean up the file and replace the currently not very nice way we decide what to build.
- [ ] Change the default `build.sh` and `build.ps1` to build in Debug mode with minimal logging, for interactive use. Update the CI builds to pass in a verbose logging flag.

### DIFF
--- a/rider/frontend.gradle
+++ b/rider/frontend.gradle
@@ -198,9 +198,13 @@ def unityeditorplugin_path = "$projectDir/../unity/build/JetBrains.Rider.Unity.E
 def unityeditorplugin_full_path = "$projectDir/../unity/build/JetBrains.Rider.Unity.Editor.Plugin.Full/Assets/Plugins/Editor/JetBrains/JetBrains.Rider.Unity.Editor.Plugin.Full.Repacked.dll"
 
 prepareSandbox {
-    // Already depends on the standard Java build/jar task
-    dependsOn buildReSharperHostPlugin, buildUnityEditorPlugin
-    dependsOn nunit // will only get run when passed RunTests
+    // Default dependsOn includes the standard Java build/jar task
+
+    if (!project.dotNetUpToDate)
+        dependsOn buildReSharperHostPlugin, buildUnityEditorPlugin
+
+    if (project.RunTests)
+        dependsOn nunit
 
     // Have dependent tasks use uptodateWhen { project.isRunningUnderCi etc. }
     //inputs.files(buildRiderPlugin.outputs)
--- a/rider/frontend.gradle
+++ b/rider/frontend.gradle
@@ -84,63 +84,6 @@ compileKotlin {
     }
 }
 
-apply plugin: 'com.jetbrains.rdgen'
-
-ext.modelDir = new File(repoRoot, "rider/protocol/src/main/kotlin/model")
-
-// TODO: Add an msbuild task for this, too?
-// Because we sometimes skip building the C# solutions (calling runIde from the C# IDE), we might
-// get into the situation where we rebuild the protocol for .kt, but not .cs
-// Maybe we should have an msbuild task to run rdgen if the input model .kt is newer than the output
-// .cs. Or make sure we don't skip the msbuild tasks if this task runs
-
-task generateRiderModel(type: RdGen) {
-    def csOutput = new File(repoRoot, "resharper/src/resharper-unity/Rider/RdUnityProtocol")
-    def ktOutput = new File(repoRoot, "rider/src/main/kotlin/com/jetbrains/rider/protocol/RdUnityProtocol")
-
-    params {
-        verbose = true
-        classpath "$rdLibDirectory/rider-model.jar"
-        sources "$modelDir/rider"
-        hashFolder = 'build/rdgen/rider'
-        packages = "model.rider"
-
-        generator {
-            language = "kotlin"
-            transform = "asis"
-            root = "com.jetbrains.rider.model.nova.ide.IdeRoot"
-            namespace = "com.jetbrains.rider.model"
-            directory = "$ktOutput"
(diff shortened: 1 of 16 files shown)


IvanPashchenko:
Looks great.

optional:
WDYT How about finally moving to kotlin from groovy for gradle? :) (not pressuring, but if you're considering rewriting `backend.gradle` maybe that's just the right time for that)?

citizenmatt:
I have no opinion either way on Kotlin/Groovy. What's the benefit? 

And are you ok with the other suggestions - mainly unzip into the gradle cache folder (which would also mean we could remove a special case for Rider in the gradle IntelliJ plugin)

IvanPashchenko:
Kotlin just has much better support in IDEA than Groovy (and the difference is only going to become bigger) which is a benefit when the build script is as large as ours. No pressure though, I can do it myself sometimes later. Just thought it was worth a mention.

Other suggestions all make sense and I agree with them.

resharper/test/src/CSharp/Feature/Services/Utils.cs
```diff
@@ -19,8 +19,10 @@ public static void CleanupOldUnityReferences(ISolution solution)
                 {
                     if (targetFrameworkScope is ProjectTargetFrameworkScope projectScope)
                     {
+#pragma warning disable 618
```
IvanPashchenko: is that's obsolete, what is the new API?

rider/buildSrc/src/main/kotlin/com/jetbrains/rider/plugins/gradle/tasks/GenerateBuildPropsTask.kt
```diff
@@ -0,0 +1,70 @@
...
+</Project>
+""")
+    }
+
+    private var parsedPackageVersion: String? = null
+
+    private fun parsePackageVersion(): String? {
+        if (parsedPackageVersion != null) return parsedPackageVersion
+
+        if (packagesDirectory == null) {
```
IvanPashchenko: is there no convenient Syntax for the asserts like in groovy? e.g. `.existsDirectory()`

rider/buildSrc/src/main/kotlin/com/jetbrains/rider/plugins/gradle/tasks/GenerateBuildPropsTask.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.TaskAction
+import java.io.File
+import java.util.regex.Pattern
+
+open class GenerateBuildPropsTask: DefaultTask() {
+
+    @Input
+    var packagesDirectory: Any? = null
```
IvanPashchenko: why `Any?` and not `File`?

rider/frontend.gradle
```diff
@@ -20,172 +30,31 @@ intellij {
...
     }
     
     intellijRepo = "https://cache-redirector.jetbrains.com/intellij-repository"
 
+    // The gradle intellij plugin has special handling for Rider to unzip the SDK into project.buildDir, because we
+    // needed to reference files from it. The ideaDependencyCachePath property allows us to put it somewhere else, and
+    // takes precedence to the Rider special handling. If we set it to an invalid path, we get the default IntelliJ SDK
+    // handling of unzipping inside the gradle cache folder. We'll prefer this to the build folder, so we don't get
+    // multiple copies if we have multiple working directories
+    // BUT... using the default gradle cache means we hit Windows path length limit. Sigh
```
IvanPashchenko: it's 2018 and we still have to deal with path length limits >_>

unity/EditorPlugin/RiderPathLocator.cs
```diff
@@ -279,6 +279,8 @@ private static string[] CollectPathsFromToolbox(string toolboxRiderRootPath, str
       return paths;
     }
 
+#pragma warning disable 0649
```
IvanPashchenko: I think it would be nice to have an explanation why the warning is disabled (I've never been to that part, so if that's obvious to anyone working on that, please ignore the request :))

rider/buildSrc/src/main/kotlin/com/jetbrains/rider/plugins/gradle/tasks/GenerateBuildPropsTask.kt
citizenmatt: Because of the type coercion (see below), you can't really assert it, you can only pass it to `project.file()` to convert it to a file and that has to be non-null.

PR #806, opened 2018-09-24
183 refactor isUnityProject properties/methods, so they do similar things on backend and frontend

van800:
+ for unity-like-folder automatically install EditorPlugin, remove Install EditorPlugin Action from Notification.
+ Avoid tracking Unity references on frontend


resharper/src/resharper-unity/Rider/UnitySolutionTracker.cs
```diff
@@ -5,40 +5,79 @@
...
-            
+            SetValues();
+
             fileSystemTracker.AdviseDirectoryChanges(lifetime, mySolution.SolutionDirectory.Combine(ProjectExtensions.AssetsFolder), false,
                 OnChangeAction);
-            fileSystemTracker.AdviseDirectoryChanges(lifetime, mySolution.SolutionDirectory.Combine(ProjectExtensions.LibraryFolder), false,
-                OnChangeAction);
-            fileSystemTracker.AdviseDirectoryChanges(lifetime, mySolution.SolutionDirectory.Combine(ProjectExtensions.ProjectSettingsFolder), false,
-                OnChangeAction);
+            fileSystemTracker.AdviseDirectoryChanges(lifetime, mySolution.SolutionDirectory.Combine(ProjectExtensions.ProjectSettingsFolder), true,
```
citizenmatt: This `true` means to watch recursively, right? Do we need it `true` for `ProjectSettings`?

citizenmatt: Ah, wait, is this so we see changes to `ProjectVersion.txt` or `*.asset`, too? If so, might be worth a comment as to why.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/DefaultRunConfigurationGenerator.kt
```diff
@@ -3,51 +3,51 @@ package com.jetbrains.rider.plugins.unity.run
...
-            override fun hasUnityReference() {
-                // Clean up the renamed "attach and play" configuration from 2018.2 EAP1-3
-                // (Was changed from a separate configuration type to just another factory under "Attach to Unity")
-                val toRemove = runManager.allSettings.filter {
-                    it.type is UnknownConfigurationType && it.name == "Attach to Unity Editor & Play"
-                }
-                for (value in toRemove) {
-                    runManager.removeConfiguration(value)
-                }
+        unityHost.model.hasUnityReference.whenTrue(componentLifetime) {
```
citizenmatt: It might be a good idea to move `unityHost.model.hasUnityReference` inside `UnityProjectDiscoverer`. That would mean `UnityProjectDiscoverer` was the one place to go to see if we're in a Unity project - either by looking at the folder structure, or based on a property from the backend.

This property is effectively the "truth" about if we're a Unity project or not. So perhaps we should look at our existing `isUnityProject` methods and make them look at `hasUnityReference`, rather than file structure? So we could have a range of functions - `isUnityProjectFolder` for when it's not loaded properly, `isUnityProjectQuick` for when it's based on folder structure (for Unity Explorer) and `isUnityProject` that also checks the backend value?

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/OpenUnityProjectAsFolderNotification.kt
```diff
@@ -76,19 +75,6 @@ class OpenUnityProjectAsFolderNotification(private val project: Project, private
...
-                        "Install Rider’s Unity Editor plugin to automatically configure Rider as the default external editor and enable advanced functionality."
-                )
-                notification.addAction(object : InstallEditorPluginAction() {
-                    override fun actionPerformed(e: AnActionEvent) {
-                        super.actionPerformed(e)
-                        notification.hideBalloon()
-                    }
-                })
-            }
-
```
citizenmatt: Why is this removed?

van800: Because we are now installing it if we detect UnityFolderStructure
https://github.com/JetBrains/resharper-unity/pull/806/files#diff-2bca04de557ea96772f6862ec6550ab4R72

resharper/src/resharper-unity/Rider/UnitySolutionTracker.cs
van800: True. Recursive for  `ProjectVersion.txt` or `*.asset`

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/DefaultRunConfigurationGenerator.kt
van800: UnityProjectDiscoverer works for Re# and Rider, while unityHost.model.hasUnityReference is only for Rider.

van800: Not sure about the second part about isUnityProjectQuick

citizenmatt: Just giving the current `isUnityProject` a name that indicates that it's based on folder structure, rather than "the truth". It _might_ be a false positive. If you gave someone the API right now, they wouldn't know the difference between `isUnityProject` and `hasUnityReference`.

citizenmatt:
LGTM!

PR #816, opened 2018-09-28
WIP, Highlighting: hot methods and invocation of costly methods

krasnotsvetov:
*TODO*


akarpov89:
Briefly looked at the files and couldn't find any highlighting tests. Please add.

resharper/src/resharper-unity/CSharp/Daemon/Stages/PerformanceCriticalCodeAnalysis/PerformanceCriticalCodeAnalysisStage.cs
```diff
@@ -0,0 +1,560 @@
...
+            #endregion
+        }
+
+        private class HotMethodAnalyzerContext
+        {
+            // Current depth of analysis
+            public int Depth { get; set; }
+            
+            // Inverted call graph for costly reachable mark propagation after visit AST
+            public IReadOnlyDictionary<IDeclaredElement, HashSet<IDeclaredElement>> InvertedCallGraph
```
akarpov89: Creating a copy of the whole dictionary isn't very good in terms of memory consumption. Maybe return the original dictionary?

akarpov89:
It's very cool to have such inspections 👍 

Though I have concerns about current implementation. In ReSharper there is a daemon stage which also builds call graph and is cross-procedural: `SynchronizationProblemAnalyzerStage`. I think `PerformanceCriticalCodeAnalysisStage` should be implemented in a similar way. I'm even wondering if it would be possible to share some code between these stages, there are certainly common parts.

While current implementation of `PerformanceCriticalCodeAnalysisStage` rebuilds the call graph on every run, `SynchronizationProblemAnalyzerStage` is incremental. Take a look how it uses `FileStructure.MembersToRehighlight` to rebuild only some parts of the graph. Also note that during the `CSharpErrorStage` `LockSectionProblemAnalyzer` saves info about presence of `lock` statement. In case there are no locks stage exists. You can actually ask Ivan Serduk for more details, he is author of the `SynchronizationProblemAnalyzerStage`.

resharper/src/resharper-unity/CSharp/Daemon/Stages/PerformanceCriticalCodeAnalysis/PerformanceCriticalCodeAnalysisStage.cs
```diff
@@ -0,0 +1,560 @@
...
+                        break;
+                    case IMethodDeclaration methodDeclaration:
+                        // check that method is hot and add it to container                        
+                        var name = methodDeclaration.DeclaredElement?.ShortName;
+                        if (name != null && ourKnownHotMonoBehaviourMethods.Contains(name))
+                            hotRootMethods.Add(methodDeclaration);                                
+                        break;
+                    case IInvocationExpression invocationExpression:
+                        // we should find 'StartCoroutine' method, because passed symbol will be hot too
+                        var reference = (invocationExpression.InvokedExpression as IReferenceExpression)?.Reference;
```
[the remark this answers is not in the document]
krasnotsvetov: InvocationExpressionReference has invalid DocumentRange (-1,-1)
(invocationExpression.InvokedExpression as IReferenceExpression)?.Reference has valid range for document

resharper/src/resharper-unity/CSharp/Daemon/Stages/PerformanceCriticalCodeAnalysis/PerformanceCriticalCodeAnalysisStage.cs
```diff
@@ -80,28 +82,111 @@ private void AnalyzeFile(ICSharpFile file, IHighlightingConsumer consumer)
...
+                var visitor = new HotMethodAnalyzer(sourceFile, consumer, ourMaxAnalysisDepth);
+                methodDeclaration.ProcessDescendants(visitor, context);
+            }
+
+            return context;
+        }
+
+        private static List<IMethodDeclaration> FindHotRootMethods([NotNull]ICSharpFile file,[NotNull] IPsiSourceFile sourceFile)
+        {
+            var result = new List<IMethodDeclaration>();
```
akarpov89: Often it's good to use `LocaList<T>` instead of `List<T>`, it doesn't allocate without need. Once you populated local list you can then invoke `.ReadOnlyList()` on it to get `IReadOnlyList<T>`.

PR #817, opened 2018-10-01
WIP: Support Unity Code Style

krasnotsvetov:
By default, Unity code style checking is enabled. It can be disabled from Unity Plugin settings


JurjenBiewenga:
I'm not quite sure what coding standard this is based on but the Unity api has a bunch of different ones, I think it would be best to have the C# standard as default.

IvanPashchenko:
@JurjenBiewenga 
This is based on what we've mostly encountered and on talks with our users/unity team. Main idea is that if user doesn't know that he can configure naming styles, he isn't annoyed with squiggles all around his code. This should eventually get replaced by auto-detecting of naming styles in R# if it is ever done, after which can safely delete this workaround.

krasnotsvetov:
see #889

PR #822, opened 2018-10-02
Added version info to preferences

### LINKED ISSUE #818: Show version of editor plugin in Unity prefs page
It can be very hard to see what version of the editor plugin is installed, especially on Mac (Windows should be able to view Properties in Explorer). It would be good if there were a text line in the Rider prefs page inside Unity that printed the version of the currently installed plugin.

JurjenBiewenga:
Fixes #818

### DIFF
--- a/unity/EditorPlugin/PluginSettings.cs
+++ b/unity/EditorPlugin/PluginSettings.cs
@@ -1,6 +1,9 @@
 using System;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
 using JetBrains.DataFlow;
 using JetBrains.Util;
 using JetBrains.Util.Logging;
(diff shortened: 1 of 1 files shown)


unity/EditorPlugin/PluginSettings.cs
```diff
@@ -318,7 +321,17 @@ private static void RiderPreferencesItem()
...
       var caption = $"<color=#0000FF>{githubRepo}</color>";
       LinkButton(caption: caption, url: githubRepo);
-
+      
+      GUILayout.FlexibleSpace();
+      GUILayout.BeginHorizontal();
+      
+      GUILayout.FlexibleSpace();
+      Assembly assembly = Assembly.GetExecutingAssembly();
+      FileVersionInfo versionInfo = FileVersionInfo.GetVersionInfo(assembly.Location);
```
citizenmatt: I think we should use `assembly.GetName().Version`. This is the ".net version" as it were, and we've had issues before where we've had the assembly version correct but not the file version.

PR #823, opened 2018-10-02
`AddLookupItems` now checks whether the item is already added

### LINKED ISSUE #685: Ensure event function completion items are distinct
E.g. completion inside `EditorWindow` shows two `Awake` methods - one from `EditorWindow` and the other from `ScriptableObject`.

JurjenBiewenga:
Fixes #685


resharper/test/data/CSharp/CodeCompletion/List/EditorWindow01.cs
```diff
@@ -0,0 +1,7 @@
+using UnityEngine;
+using UnityEditor;  // More types in scope
+
+public class A : EditorWindow
```
citizenmatt: Just add a quick comment here that this class inherits `Awake` from both `EditorWindow` and `ScriptableObject`.

PR #834, opened 2018-10-08
fix compile options pages

van800:


rider/src/main/kotlin/com/jetbrains/rider/settings/UnityPluginOptionsPage.kt
```diff
@@ -2,4 +2,8 @@ package com.jetbrains.rider.settings
 
 import com.jetbrains.rider.settings.simple.SimpleOptionsPage
 
-class UnityPluginOptionsPage : SimpleOptionsPage("Unity Engine", "UnityPluginSettings")
\ No newline at end of file
+class UnityPluginOptionsPage : SimpleOptionsPage("Unity Engine", "UnityPluginSettings") {
+    override fun getId(): String {
```
IvanPashchenko: fun note: can be written as 
`override fun getId(): String = "preferences.build.unityPlugin"`
Kotlin :)

PR #835, opened 2018-10-08
Feature/component references

citizenmatt:
Adds references to the types in the string literal arguments to:

* `GameObject.AddComponent`
* `GameObject.GetComponent`
* `ScriptableObject.CreateInstance`

Resolves:

* Namespace qualified types. E.g. `GetComponent("UnityEngine.Grid")`
* Type in any namespace, by type name alone. E.g. `GetComponent("Grid")`
* Unresolved types or namespaces are highlighted as resolve errors
* Multiple candidates are highlighted as warning, with quick fix to use the correct type
* Types with an incorrect base type are highlighted with a warning. User code must be derived from `MonoBehaviour`, built in code derives from `Component`. Also checks base type for `ScriptableObject`.
* Types with generic type parameters are highlighted as a warning (Unity doesn't support type parameters on `MonoBehaviour` or `ScriptableObject` at all)

Code completion:

* Namespaces and types in the global namespace
* Namespaces and types qualified by preceding namespace
* Only public types are shown in code completion, but resolve will match any visibility
* Does NOT complete types from any namespace, e.g. `GetComponent("Grid")` will resolve correctly to `UnityEngine.Grid`, but won't provide any code completion

Inspections:

* Updates `PreferGenericMethodOverloadAnlayzer` to get type information by resolving the references. This removes a couple of highlights, because they're now handled by resolve.


resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/PreferGenericMethodOverloadAnalyzer.cs
```diff
@@ -1,202 +1,76 @@
...
-            
-            var typeNode = factory.CreateTypeUsageNode(literal, context);
-            var typeReference = ((IReferenceName)typeNode.FirstChild)?.Reference;
-            if (typeReference == null)
+            // Don't add anything unless ALL references resolve properly
+            foreach (var reference in references)
             {
-                return Array.Empty<ITypeElement>();
+                var resolveInfo = reference.Resolve();
+                if (!resolveInfo.Info.ResolveErrorType.IsAcceptable)
```
krasnotsvetov: In this case, we will never show quick fix with possibility to choose specified class in case with ambiguous classes
````
namespace A 
{
    class Test  : MonoBehaviour
    {
    }
}

namespace B
{
   class Test : MonoBehaviour 
   {
   }
}

class Example : MonoBehaviour 
{
    void Start() 
    {
        var test = GetComponent("Test");
    }
}

```

resharper/src/resharper-unity/CSharp/Feature/Services/Daemon/UnityObjectIncorrectBaseTypeErrorHandler.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace JetBrains.ReSharper.Plugins.Unity.CSharp.Feature.Services.Daemon
+{
+    [Language(typeof(CSharpLanguage))]
+    public class UnityObjectIncorrectBaseTypeErrorHandler : IResolveProblemHighlighter
+    {
+        // We only get called for our known resolve errors
+        public IHighlighting Run(IReference reference)
+        {
+            var errorType = reference.CheckResolveResult();
+            if (errorType == UnityResolveErrorType.EXPECTED_COMPONENT)
```
krasnotsvetov: I don't see any warning on my Mac OS machine, only highlighting (no tooltip). I will check it tomorrow on Windows machine.
In inspection result the correct message is presented.

resharper/src/resharper-unity/CSharp/Feature/Services/QuickFixes/PreferGenericMethodOverloadQuickFix.cs
```diff
@@ -34,40 +17,27 @@ namespace JetBrains.ReSharper.Plugins.Unity.CSharp.Feature.Services.QuickFixes
     [QuickFix]
     public class PreferGenericMethodOverloadQuickFix : IQuickFix
     {
-        [NotNull] public static readonly InvisibleAnchor IntentionAnchor =
-            new InvisibleAnchor(IntentionsAnchors.QuickFixesAnchor);
-
         private readonly IInvocationExpression myInvocationExpression;
         private readonly string myMethodName;
-        private readonly ITypeElement[] myAvailableTypes;
+        private readonly ITypeElement myTypeElement;
```
krasnotsvetov: I think that we should return previous behaviour with possibility to choose.

resharper/test/data/CSharp/Resolve/UnityObjectTypeInStringLiteral/CodeCompletion/AddComponent01.cs
```diff
@@ -0,0 +1,9 @@
+using UnityEngine;
+
+public class Test
+{
+    public void Method(GameObject go)
+    {
+        go.AddComponent("{caret}");
+    }
+}
```
krasnotsvetov:
```
namespace A
{
	public class Test2 : MonoBehaviour
	{
				
	}
}

public class Test : MonoBehaviour
{
	void Update ()
	{
			Debug.Log(GetComponent("Tes{caret}")); 
	}
}
```
Code completion do not suggest `Test2`

resharper/src/resharper-unity/CSharp/Daemon/Stages/Analysis/PreferGenericMethodOverloadAnalyzer.cs
citizenmatt: The reference resolve will show a warning, with a quick fix to choose which one. Selecting that will change it to a fully qualified string, and then we get the quick fix to rewrite as generic method. I tried it first with the quick fix on multiple candidates, but the interaction is weird - the warning for the string literal overwrites the ambiguous match warning, which I think is more important.

resharper/test/data/CSharp/Resolve/UnityObjectTypeInStringLiteral/CodeCompletion/AddComponent01.cs
citizenmatt: No, it can't. As mentioned in the PR details above. We can't provide code completion for ALL types, the list is too big and too full of items that are irrelevant.

PR #838, opened 2018-10-09
More debugger extensions

citizenmatt:
Adds extra nodes when debugging Unity:

* Adds `Components` group as child of an instance of `GameObject`. Lists the components that are attached to that game object.
* Adds `Children` group to an instance of `GameObject`. Lists the child game objects as per the Hierarchy tool window in the Unity editor.
* Adds `Game Objects` group to an instance of `Scene`. Lists the root game objects of a scene, as per the Hierarchy tool window. And because each `GameObject` now has a `Children` node, you can examine the entire hierarchy of a scene.

Also refactors the debugger extension. Most major change is moving it to a top level `debugger` folder, rather than have it mixed in with the backend plugin.


citizenmatt:
OK, thanks. I'll merge this for now, and I've created #840 to track disabling this. I'll also follow up with you on options for how we can do this.

PR #843, opened 2018-10-11
improve adding xcode dll reference to csproj

van800:


Ethan-VisualVocal:
Note: When upgrading from an existing Rider 2018.2 (.csproj and .sln files exist), Rider 2018.3 EAP Build #RD-183.3647.257 (Resharper-Unity plugin v2018.3.0.26) doesn't correctly modify projects to have these references.

However, if I close down Unity the new Resharper-Unity v2018.3.0.26 plugin gets installed, and delete all Unity auto-generated files (including .csproj and .sln files), then reopen Unity and then reopen Rider 2018.3 EAP, it DOES correctly reference the Xcode dlls.

So, it works, it just doesn't have a smooth upgrade path.  Let me know if I can provide more info!

Other system info: 
- Unity 2017.4.12f1
- macOS 10.13.6

image

van800:
@Ethan-VisualVocal Thanks! I added a new issue for the case. https://github.com/JetBrains/resharper-unity/issues/855

PR #845, opened 2018-10-13
Add StartUnityAction

van800:
New action starts Unity and in Unity it automatically starts new Rider


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/actions/StartUnityAction.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.project.DumbAwareAction
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.SystemInfo
+import com.jetbrains.rider.plugins.unity.util.UnityIcons
+import com.jetbrains.rider.plugins.unity.util.UnityInstallationFinder
+import com.microsoft.alm.helpers.Path
+
+
+open class StartUnityAction : DumbAwareAction("Start Unity", "Start Unity with current project.", UnityIcons.Actions.ImportantActions) {
```
citizenmatt: Description doesn't normal end with a period '.'

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/actions/StartUnityAction.kt
```diff
@@ -0,0 +1,33 @@
...
+    companion object {
+         fun StartUnity(project: Project, vararg args: String) {
+             val appPath = UnityInstallationFinder.getInstance(project).getApplicationPath()
+             if (appPath == null) return
+
+             var path = appPath.toString()
+             if (SystemInfo.isMac)
+                 path = Path.combine(path, "Contents/MacOS/Unity")
+             val processBuilderArgs = mutableListOf<String>(path, "-projectPath", project.basePath.toString())
+             args.forEach { processBuilderArgs.add(it)}
```
citizenmatt: I think you can do something like `processBuilderArgs.addAll(*args)`

unity/EditorPlugin/PluginSettings.cs
```diff
@@ -25,6 +23,37 @@ public enum AssemblyReloadSettings
     StopPlayingAndRecompile = 2
   }
 
+  // this is used to start Rider automatically via Unity commandline switches
```
citizenmatt: Pull the new class out into a new file?

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/OpenUnityProjectAsFolderNotification.kt
```diff
@@ -49,6 +57,9 @@ class OpenUnityProjectAsFolderNotification(private val project: Project, private
                     ProjectUtil.closeAndDispose(project)
                     WelcomeFrame.showIfNoProjectOpened()
                 }
+                if (hyperlinkEvent.description == "reopen") {
+                    StartUnityAction.StartUnity(project, "-executeMethod", "JetBrains.Rider.Unity.Editor.RiderMenu.MenuOpenProject")
```
citizenmatt: I think move `-executeMethod` and static method arguments into the `StartUnity` method. Then `StartUnity` is the only place that has to know about Unity's command line args. Perhaps even rename to something like `OpenUnityProject` or `OpenProjectInUnity`. If we need to reuse this in some other place with other parameters or static method names, we can refactor at that time.

citizenmatt: What happens if the plugin isn't installed? Is there an error?

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/OpenUnityProjectAsFolderNotification.kt
```diff
@@ -31,14 +34,19 @@ class OpenUnityProjectAsFolderNotification(private val project: Project, private
...
+var reopenText = "<a href=\"reopen\">reopen</a>"
+            val (status, editorInstanceJson) = EditorInstanceJson.load(project)
+            if (status == EditorInstanceJsonStatus.Valid) {
+                 // do not show Action
+                reopenText = "reopen"
+            }
             val content = if (solutionDescription.projectFilePaths.isEmpty()) {
                 "This looks like a Unity project. C# and Unity specific functionality is not available when the project is opened as a folder." +
-                    " Please <a href=\"close\">close</a> and reopen through the Unity editor, or by opening a .sln file."
+                    " Please <a href=\"close\">close</a> and "+reopenText+" through the Unity editor, or by opening a .sln file."
```
citizenmatt: We can only click on one link before the notification is gone, so it's going to be hard for the user to choose between "close" and "reopen".

So if we have `EditorInstance.json` (Unity is currently running), leave it as it was:

```
Please <a href=\"close\">close</a> and reopen through the Unity editor, or by opening a .sln file.
```

But if we don't have it (Unity isn't running), we should probably only have one action link:

```
Please <a href=\"reopen\">click here</a> to start Unity, generate a solution file and reopen the project.
```

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/OpenUnityProjectAsFolderNotification.kt
citizenmatt: Should this also close the current project?

van800: Plugin is expected to be automatically installed by the same rules, which show this notification.

van800:
@citizenmatt fixed everything, thank you very much. Now I like, how it behaves.

Issue #846, opened 2018-10-14
Add action to force Unity AppDomain Reload, even when none of the scripts was changed


van800:
One of the participants on GIC mentioned that he is doing frequently different [InitializeOnLoad] tasks.
Debugging it requires forcing Unity to do AppDomainReload.
a) Just Entering Play is not good, because some of the projects have significant delays on that.
b) SyncSolution, which RIder is calling doesn't do anything if there were no file-changes. (SyncSolution calls AssetDatabase.Refresh internally)

For a workaround he is doing Reimport of one script. But this can be done only in Unity itself.

PR #847, opened 2018-10-15
call RequestScriptReload on force Refresh

van800:
https://github.com/JetBrains/resharper-unity/issues/846


citizenmatt:
I don't think we should be causing an AppDomain reload every time we click the refresh button. Should we create a new action instead?

van800:
screen shot 2018-10-16 at 9 06 49 am
Initial use-case for "Trigger Refresh in Unity Editor" was the case when "Automatically Refresh" is disabled in Rider Settings but explicit pressing button/calling shortcut still works.
I do not think we should add a new Action.
**User is explicitly calling Action via button or shortcut means he wants the Refresh/Reload, so it seems logical to force it.**
It will widen the initial use-case, but I can't see why it might be bad thing.

StephenHodgson:
I think as long as people know exactly what it does, it should be okay.

PR #869, opened 2018-10-30
Fix warnings in frontend + enable warnings as errors

citizenmatt:
Also deletes some unused action group code and registration


IvanPashchenko:
Merging as this is green on TC

PR #870, opened 2018-10-30
YAML PSI merge

citizenmatt:
Import the YAML PSI from [citizenmatt/resharper-yaml](https://github.com/citizenmatt/resharper-yaml) and build with the main solution. It is not wired up to any file types, so essentially does nothing apart from building and bumping up the number of tests.

The main reason for this PR is to get the massive import out of the way so that subsequent PRs don't include all of the YAML PSI files and are easier to review.


resharper/resharper-yaml/src/Daemon/Stages/YamlDaemonStageBase.cs
```diff
@@ -0,0 +1,48 @@
...
+      return process.SourceFile.GetPsiFiles<YamlLanguage>()
+        .SelectNotNull(file => CreateProcess(process, settings, processKind, (IYamlFile) file));
+    }
+
+
+    protected abstract IDaemonStageProcess CreateProcess(IDaemonProcess process, IContextBoundSettingsStore settings,
+      DaemonProcessKind processKind, IYamlFile file);
+
+    public ErrorStripeRequest NeedsErrorStripe(IPsiSourceFile sourceFile, IContextBoundSettingsStore settingsStore)
+    {
```
IvanPashchenko: What's this for and why is this method required?

resharper/resharper-yaml/src/Psi/Parsing/YamlLexerGenerated.cs
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    private TokenNodeType HandleImplicitKey()
+    {
+      parentNodeIndent = yy_buffer_start - lastNewLineOffset;
+
+      // We special case implicit keys as a single line ns-plain followed by a colon
+      // But we want to return that as separate tokens. So match it, rewind and let
+      // the next pass handle the whitespace and colon
+      // TODO: This doesn't reset line indents or last new line offset!!
```
IvanPashchenko: looks like important todo :)

resharper/resharper-yaml/src/Psi/Parsing/YamlTokenType.CommentNodeType.cs
```diff
@@ -0,0 +1,35 @@
...
+      {
+        return new Comment(buffer.GetText(new TextRange(startOffset.Offset, endOffset.Offset)));
+      }
+
+      public override LeafElementBase Create(string token)
+      {
+        return new Comment(token);
+      }
+
+      // NOTE: Not filtered
```
IvanPashchenko: why?

resharper/resharper-yaml/src/Psi/YamlProjectFileLanguageService.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    public override ILexerFactory GetMixedLexerFactory(ISolution solution, IBuffer buffer, IPsiSourceFile sourceFile = null)
+    {
+      var languageService = YamlLanguage.Instance.LanguageService();
+      return languageService?.GetPrimaryLexerFactory();
+    }
+
+    protected override PsiLanguageType PsiLanguageType => (PsiLanguageType) YamlLanguage.Instance ?? UnknownLanguage.Instance;
+
+    // TODO: Proper icon!
```
IvanPashchenko: do we have a request for Kostya?

IvanPashchenko:
Are you sure that we need to bind to `.yaml` extension? Why not `.meta` or `.unity`?
Will this parser launch in non-Unity solutions?

resharper/resharper-yaml/.editorconfig
```diff
@@ -0,0 +1,8 @@
+root = true
```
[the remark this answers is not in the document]
citizenmatt: We already have one at the root of the main solution. There's a duplicate one here because this subfolder is also the citizenmatt/resharper-yaml repo (at least for now, so I can do a standalone R# plugin). There is a problem that I should have fixed before merging - the yaml code is formatted at 2 spaces, while the unity one is at 4 :) I'll do a separate PR fixing that sometime later.

resharper/resharper-yaml/src/Daemon/Stages/YamlDaemonStageBase.cs
citizenmatt: The perils of an initial commit style PR :) This method used to be on `IDaemonStage`, but has been removed. I've already removed it in following commits.

resharper/resharper-yaml/src/Psi/Parsing/YamlLexerGenerated.cs
citizenmatt: They're all important :) I agree, but I haven't hit it in any tests yet. I'll get round to it before RTM.

resharper/resharper-yaml/src/Psi/Parsing/YamlTokenType.CommentNodeType.cs
citizenmatt: Because the spec only allows comments in certain places, so we want the parser to see them. I'll add a comment.

resharper/resharper-yaml/src/Psi/YamlProjectFileLanguageService.cs
citizenmatt: https://youtrack.jetbrains.com/issue/RIDER-21140

citizenmatt:
With this PR, yes, it's bound to just `.yaml`. Following commits change this. For Rider, it doesn't handle `.yaml`, but does handle `.unity`, `.meta` and `.asset`. This PR is the initial commit "add yaml PSI and get it building" commit. The following PR (that should have less than 356 files 😄) fixes all this. I won't merge this until both are ready.

PR #871, opened 2018-10-30
Fix C# language level override with "latest"

citizenmatt:
It was incorrectly treating this as though it were "default", and trying to apply heuristics.  It now correctly treats this as "latest", e.g. C# 7.3 at time of going to press.

### DIFF
--- a/resharper/resharper-unity/src/ProjectModel/Caches/UnityProjectFileCacheProvider.cs
+++ b/resharper/resharper-unity/src/ProjectModel/Caches/UnityProjectFileCacheProvider.cs
@@ -101,19 +101,18 @@ namespace JetBrains.ReSharper.Plugins.Unity.ProjectModel.Caches
                 if (xmlElement == null)
                     continue;
 
+                // We can't just grab the value here because there may be multiple values set, one per configuration.
+                // I haven't seen Unity or Rider do this, so it must be VSTU, but I don't have proof...
                 if (xmlElement.GetElementsByTagName("LangVersion").Count > 0)
-                {
                     explicitLangVersion = true;
-                }
 
-                if (versionFromDll != null) 
+                if (versionFromDll != null)
                     continue;
-                
-                // Ideally, we could get the defines through the project model
-                // (see IManagedProjectConfiguration), but that only seems to
-                // give us the currently active project settings, and Unity's
-                // own .csproj creator only sets the version for the Debug build,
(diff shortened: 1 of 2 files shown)


citizenmatt:
The Travis builds completed successfully, but didn't update their statuses. Oh well. Merging!

benoit-dubreuil:
It still happens in the latest Rider 2020.2 version. I accidentally clicked on `reset` and can't undo that, so Rider would automatically select the right C# version. So, when I try to select C# version 7.3, it automatically switches to the `C# latest minor (latest)` either at the same time or a few seconds / minutes later (try typing analyzing the code or typing a line or two).

It is frustrating, as it suggests code changes related to C# 8.0 and this version isn't supported in Unity.

van800:
What is your setup (Unity version, Rider version, OS)?
Please attach logs from Help->Collect Logs.
If you have a *.dotSettings file near sln, please attach it also.
@benoit-dubreuil

benoit-dubreuil:
Unity : 2019.4.3f1
Rider : 2020.2
OS : Windows 10 Home - 64-bit

[rider-logs-20200817-135947.zip](https://github.com/JetBrains/resharper-unity/files/5086005/rider-logs-20200817-135947.zip)

[dotSettings.zip](https://github.com/JetBrains/resharper-unity/files/5086021/dotSettings.zip)

I noticed my problem (the consequence, not the cause) is in the file `EditMode.csproj.DotSettings`, more specifically the line :
`<s:String x:Key="/Default/CodeInspection/CSharpLanguageProject/LanguageLevel/@EntryValue">Default</s:String></wpf:ResourceDictionary>`

I deleted that file and it fixed my problem but it's still a bug as far as I know.

van800:
I wonder how that setting was introduced. Do you remember adding it? Is it in your vcs or only locally?

benoit-dubreuil:
> I wonder how that setting was introduced.
Step1
Step2

> Do you remember adding it?
Yes. I clicked on the `reset` button shown on the previous image because I was testing C# 8.0 features.

> Is it in your vcs or only locally?
Only locally.

van800:
Thanks! I have started internal discussion https://jetbrains.slack.com/archives/C53QFRZ7G/p1597695558405100

van800:
Request opened https://youtrack.jetbrains.com/issue/RIDER-49393

PR #878, opened 2018-11-02
Updates in performance highlightings.

krasnotsvetov:
1.  Add context "Move to start", "Move to awake" and "Move outside the loop" if current operation in loop. (All context actions is available only in per frame methods and methods that are reachable from them)
 a. Currently not solution-wide (only for opened file)
 b. This operations can break semantics, so they are just only context actions, not quickfixes
2. Add highlighting on gutter for each performance method
3. Add possibility to disable highlightings


resharper/resharper-unity/src/Settings/PerProjectSettings.cs
```diff
@@ -170,13 +170,15 @@ private void InitLanguageLevelSettings(IProject project, SettingsStorageMountPoi
             // * Unity3dRider can set `TargetFrameworkVersion` to `v4.5` on non-Windows machines to fix
             //   an issue resolving System.Linq
 
-            var languageLevel = ReSharperSettingsCSharpLanguageLevel.Default;
+            var languageLevel = ReSharperSettingsCSharpLanguageLevel.CSharp40;
```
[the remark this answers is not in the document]
krasnotsvetov: Out-of-date 183's fixes. Fixed.

resharper/resharper-unity/src/CSharp/Daemon/Stages/Analysis/StringBasedMethodInvocationProblemAnalyzer.cs
```diff
@@ -41,7 +42,11 @@ protected override void Analyze(IInvocationExpression expression, ElementProblem
             if (info.ResolveErrorType != ResolveErrorType.OK)
                 return;
 
-            var method = (info.DeclaredElement as IMethod).NotNull("info.DeclaredElement as IMethod != null");
+
+            var method = info.DeclaredElement as IMethod; // e.g. localfunction declared element, (when removing `}` and method below is parsed like local function)
```
TessenR: The comment looks strange to me. 
`IInvocationExpression` can be just a local function invocation in a correctly parsed code.
The reference expression can also resolve to almost anything e.g. fields, properties, local variables, parameters etc if they are delegates.
E.g. 
```
Action a = () => { };
a();
```

Why mention parsing recover strategy here?

resharper/resharper-unity/src/CSharp/Daemon/Stages/PerformanceCriticalCodeAnalysis/PerformanceCriticalCodeAnalysisStage.cs
```diff
@@ -182,9 +222,12 @@ private static HashSet<IMethodDeclaration> FindHotRootMethods([NotNull]ICSharpFi
 
                         break;
                     case IMethodDeclaration methodDeclaration:
-                        // check that method is hot and add it to container
```
TessenR: It looks like the code above skips types nested in classlike declarations not inherited from `MonoBehaviour` which doesn't seem intended
e.g.
```cs
class C 
{
  class NotProcessed : MonoBehaviour { ... }
}
```

resharper/resharper-unity/src/CSharp/Daemon/Stages/PerformanceCriticalCodeAnalysis/PerformanceCriticalCodeAnalysisStage.cs
```diff
@@ -250,87 +293,78 @@ private static IMethod ExtractMethodDeclarationFromStartCoroutine([NotNull]ICSha
...
             {
                 mySourceFile = sourceFile;
                 myConsumer = consumer;
+                myCheckForInterrupt = checkForInterrupt;
                 myMaxDepth = maxDepth;
             }
 
+            public override void VisitMethodDeclaration(IMethodDeclaration methodDeclarationParam, HotMethodAnalyzerContext context)
+            {
+                context.MarkCurrentAsVisited();
```
TessenR: It looks like partial methods will have one random declaration processed (probably not even containing the actual code)

resharper/resharper-unity/src/CSharp/Daemon/Stages/PerformanceCriticalCodeAnalysis/PerformanceCriticalCodeAnalysisStage.cs
```diff
@@ -250,87 +293,78 @@ private static IMethod ExtractMethodDeclarationFromStartCoroutine([NotNull]ICSha
...
                 if (declaredElement == null)
                     return;
                 
-                context.RegisterInvocationInMethod(declaredElement, reference);
-                context.MarkCurrentAsVisited();
+                context.RegisterInvocationInMethod(declaredElement, invocationExpressionParam);
 
                 // find all declarations in current file
-                var declarations = declaredElement.GetDeclarationsIn(mySourceFile).Where(t => t.GetSourceFile() == mySourceFile);
+                var declaration = declaredElement.GetDeclarationsIn(mySourceFile).FirstOrDefault(t => t.GetSourceFile() == mySourceFile);
```
TessenR: `find all declarations in current file`
`.FirstOrDefault(...);`

It looks like the comment is outdated.

I wonder why only the first declaration is processed though. It's possible to have multiple declarations of the same method with partial types. E.g.
```cs
partial class P
{
  partial void M();
}

partial class P
{
  partial void M() => Console.WriteLine();
}
```

Current code will process one random possibly empty declaration from the code above.

TessenR: Regarding `GetDeclarationsIn(mySourceFile)` call which returns declarations from other files - it looks like a copy-paste mistake in the source code for `CSharpMethod`. I'd recommend fixing the root problem instead of hacking it multiple times here.

resharper/resharper-unity/src/CSharp/Daemon/Stages/PerformanceCriticalCodeAnalysis/PerformanceCriticalCodeAnalysisStage.cs
```diff
@@ -359,7 +412,10 @@ public override void VisitEqualityExpression(IEqualityExpression equalityExpress
 
             public bool InteriorShouldBeProcessed(ITreeNode element, HotMethodAnalyzerContext context)
             {
-                return !(element is ICSharpClosure);
+                if (element is ICSharpClosure closure)
```
TessenR: It looks like this code should process local functions when their invocations are found.
Is it intended to ignore costly calls within lambda / anonymous method / query expressions? 
e.g.
```cs
void M()
{
  Action a = () => CostlyInvocation();
  a();
}
```

TessenR: I haven't seen any code that will mark local function invocations (or any closures, altough it seems that any othe forms of closures aren't processed) as visited as the `MarkCurrentAsVisited` method is only called within `VisitMethodDeclaration`. Therefore they'll probably be analyzed multiple times. While it won't cause stackoverflow exception due to the depth limit it can cause a huge performance hit especially in case of large recursive functions since they will be analyzed `MaxDepth` times for each call. Please check performance for code like this:

```cs
void M()
{
  Trigger(); // start inspecting the local function

  void Trigger()
  {
    Local(); // inspect the second local function
    Trigger(); // inspect Trigger and consequently Local again MaxDepth times
  }

  void Local()
  {
    // 1000+ lines of code here, inspect them all
    Local(); // inspect Local MaxDepth times recursively
  }
}
```

Will this analyze the `Local` source `MaxDepth`^2 times?

resharper/resharper-unity/src/CSharp/Feature/Services/QuickFixes/MoveQuickFixes/AbstractMoveQuickFix.cs
```diff
@@ -0,0 +1,151 @@
...
+            var result = new List<IntentionAction>();
+            if (MonoBehaviourMoveUtil.IsExpressionAccessibleInScript(ToMove))
+            {
+                result.Add(new IntentionAction(new MoveAction(MonoBehaviourScript, ToMove, "Start", FieldName), BulbThemedIcons.ContextAction.Id, IntentionsAnchors.ContextActionsAnchor ));
+                result.Add(new IntentionAction(new MoveAction(MonoBehaviourScript, ToMove, "Awake", FieldName), BulbThemedIcons.ContextAction.Id, IntentionsAnchors.ContextActionsAnchor ));
+            }
+
+            var loopAction = TryCreateMoveFromLoopAction(ToMove);
+            if (loopAction != null)
+                result.Add(loopAction);
```
TessenR: I wonder if actions for intermediate loops will be useful if nested under the action for a topmost loop but it'd probably be hard to come up with good unique captions for them.

resharper/resharper-unity/src/CSharp/Feature/Services/QuickFixes/MoveQuickFixes/AbstractMoveQuickFix.cs
```diff
@@ -0,0 +1,151 @@
...
+
+            return result;
+        }
+
+        private IntentionAction TryCreateMoveFromLoopAction([NotNull] ICSharpExpression toMove)
+        {
+            ITreeNode currentExpression = toMove;
+            ILoopStatement previousLoop = null;
+
+            int closureStartOffset = currentExpression.GetContainingNode<ICSharpClosure>()?.GetTreeStartOffset().Offset ?? 0;
```
TessenR: I believe something like this will be shorter and more expressive than offset checks. It also does not require multiple traversals of containing nodes in order to find out if there is a closure among them at all.
```cs
foreach (var node in currentExpression.ContainingNodes())
{
  if (node is ICSharpClosure) break;
  if (node is ILoopStatement loop)
  {
    if (IsAvailableToMoveFromLoop())
      previousLoop = loop;
    else
     break;
  }
}
```

resharper/resharper-unity/src/CSharp/Feature/Services/QuickFixes/MoveQuickFixes/AbstractMoveQuickFix.cs
```diff
@@ -0,0 +1,151 @@
...
+            protected override Action<ITextControl> ExecutePsiTransaction(ISolution solution, IProgressIndicator progress)
+            {
+                var anchor = myLoopStatement as ICSharpStatement;
+
+                var declaredElement = MonoBehaviourMoveUtil.GetDeclaredElementFromParentDeclaration(myToMove);
+                var baseName = myVariableName ?? MonoBehaviourMoveUtil.CreateBaseName(myToMove, declaredElement);
+                var name = NamingUtil.GetUniqueName(myToMove, baseName, NamedElementKinds.Locals, de => !de.Equals(declaredElement));
+                
+                var factory = CSharpElementFactory.GetInstance(myToMove);
+                var originMyToMove = myToMove.Copy();
```
TessenR: `.Copy()` is not the best method for moving tree nodes, try `.CopyWithResolve()` instead.
The difference between `.Copy()` and `.CopyWithResolve()` is that `.CopyWithResolve()` stores original resolve results in the copied non-physical tree which are then used by tree modification to rebind references at their new location. Generally you'd almost always prefer `.CopyWithResolve()` unless it's really just a text modification

Here are some examples where it works differently:
```cs
class D
{
  public void Source()
  {
    M(); // move this
  }

  public void M()
  {
    // to this location
    // .CopyWithResolve result: this.M();
    // .Copy result: M();
    void M() { }
  }
}
```

```cs
using System;

namespace N
{
  using static S1;

  partial class C
  {
    public void M1()
    {
      Method(); // copy this
    }
  }
}

namespace N
{
  using static S2;

  partial class C
  {
    public void M2()
    {
      // to this
      // .CopyWithResolve result: S1.Method();
    }
  }
}

static class S1
{
  public static extern void Method();
}

static class S2
{
  public static extern void Method();
}
```

resharper/resharper-unity/src/CSharp/Feature/Services/QuickFixes/MoveQuickFixes/MonoBehaviourMoveUtil.cs
```diff
@@ -0,0 +1,263 @@
...
+namespace JetBrains.ReSharper.Plugins.Unity.CSharp.Feature.Services.QuickFixes.MoveQuickFixes
+{
+    internal static class MonoBehaviourMoveUtil
+    {
+        [CanBeNull]
+        public static IMethodDeclaration GetMonoBehaviourMethod([NotNull] IClassDeclaration classDeclaration, [NotNull] string name)
+        {
+            classDeclaration.GetPsiServices().Locks.AssertReadAccessAllowed();
+            
+            return classDeclaration.MethodDeclarations.FirstOrDefault(t => t.NameIdentifier.Name.Equals(name));
```
TessenR: I'd assert there is at most one such declaration if it's not expected to have partial methos here.
Or use `.SingleItem(() => ...)` to not handle invalid code with multiple declarations.

resharper/resharper-unity/src/CSharp/Feature/Services/QuickFixes/MoveQuickFixes/MonoBehaviourMoveUtil.cs
```diff
@@ -0,0 +1,263 @@
...
+            toMove.GetPsiServices().Locks.AssertReadAccessAllowed();
+
+            var sourceFile = toMove.GetSourceFile();
+            if (sourceFile == null)
+                return false;
+            
+            var loopStartOffset = loop.GetTreeStartOffset().Offset;
+            return IsAvailableToMoveInner(toMove, declaredElement => 
+                    declaredElement.GetDeclarationsIn(sourceFile)
+                    .FirstOrDefault(declaration => declaration.GetSourceFile() == sourceFile)?.GetTreeStartOffset().Offset < loopStartOffset
```
TessenR: It looks like this `isElementIgnored` lambda will disallow the following transformation:
```cs
for (;;)
{
  M(Local()); // move me outside of the loop
}

void Local() { }
```
just because the local function is declared after the loop.

Wouldn't it be better to check that the declaration is not _within_ the loop instead of _before_ the loop?

JFYI: this will still break compilation in case local variables are declared but not definitely assigned before the loop. However it seems that correct check for this might be too expensive for CA/QF availability.

resharper/resharper-unity/src/CSharp/Feature/Services/QuickFixes/MoveQuickFixes/MonoBehaviourMoveUtil.cs
```diff
@@ -0,0 +1,263 @@
...
+        private static bool IsAvailableToMoveInner([NotNull] ITreeNode toMove, [CanBeNull] Func<IDeclaredElement, bool> isElementIgnored = null)
+        {
+            var nodeEnumerator = toMove.ThisAndDescendants();
+            while (nodeEnumerator.MoveNext())
+            {
+                var current = nodeEnumerator.Current;
+                IDeclaredElement declaredElement;
+                switch (current)
+                {
+                    case IDeclaration declaration:
```
TessenR: It looks like the current code will disallow the following based on the local variable declaration presence:
`M(() => { var x = 0; });`
This doesn't look intended, I guess it shouldn't process declarations in nested scopes. 

If this declaration check should disallow code like this: `M(out var x)` there's an existing extension method `ICSharpExpression.GetDeclaredExpressionVariables` to collect such variables which does not traverse closures.

resharper/resharper-unity/src/CSharp/Feature/Services/QuickFixes/MoveQuickFixes/MonoBehaviourMoveUtil.cs
```diff
@@ -0,0 +1,263 @@
...
+                        declaredElement = declaration.DeclaredElement;
+                        break;
+                    case IReferenceExpression referenceExpression:
```
TessenR: Why only `IReferenceExpression` is checked?

E.g. `T` is not a reference expression here:
```cs
  public void M<T>()
  {
    GetComponent<T>();
  }
```

`node.GetReferences()` seems like a better choice here.

resharper/resharper-unity/src/CSharp/Feature/Services/QuickFixes/MoveQuickFixes/MonoBehaviourMoveUtil.cs
```diff
@@ -0,0 +1,263 @@
...
+                        break;
+                }
+                
+                if (declaredElement == null || isElementIgnored != null && isElementIgnored(declaredElement))
+                    continue;
+
+                // if declared element is local, we can't move our expression outside the method
+                switch (declaredElement)
+                {
+                    case ILocalVariableDeclaration _:
```
TessenR: Please do not check declared element types for declaration types.
While they sometimes share the actual implementation type this is just an implementation detail and should never be used.

In the current code e.g. `IParameterDeclaration` is unlikely to share the same implementation type as `IParameter`

resharper/resharper-unity/src/CSharp/Feature/Services/QuickFixes/MoveQuickFixes/MonoBehaviourMoveUtil.cs
```diff
@@ -0,0 +1,263 @@
...
+
+            if (!isVoid)
+            {
+                var field = factory.CreateFieldDeclaration(type, name);
+                field.SetAccessRights(AccessRights.PRIVATE);
+
+                classDeclaration.AddClassMemberDeclaration(field);
+            }
+
+            var initialization = isVoid ? factory.CreateStatement("$1;", name, expression.Copy()) :
```
TessenR: `.Copy()`

resharper/resharper-unity/src/CSharp/Feature/Services/QuickFixes/MoveQuickFixes/MonoBehaviourMoveUtil.cs
```diff
@@ -0,0 +1,263 @@
...
+                var field = factory.CreateFieldDeclaration(type, name);
+                field.SetAccessRights(AccessRights.PRIVATE);
+
+                classDeclaration.AddClassMemberDeclaration(field);
+            }
+
+            var initialization = isVoid ? factory.CreateStatement("$1;", name, expression.Copy()) : 
+                factory.CreateStatement("$0 = $1;", name, expression.Copy());
+            
+            var body = methodDeclaration.EnsureStatementMemberBody();
```
TessenR: I wonder how this would/should work with code like this:

```cs
class MonoBehaviour
{
  public virtual extern void Start();
}

abstract class C : MonoBehaviour
{
  public abstract override void Start(); // move something to this method
}
```

resharper/resharper-unity/src/CSharp/Feature/Services/QuickFixes/MoveQuickFixes/MonoBehaviourMoveUtil.cs
```diff
@@ -0,0 +1,263 @@
...
+        /// <summary>
+        /// If current expression is used as initializer for local variable, declared element for this variable will be returned
+        /// </summary>
+        public static IDeclaredElement GetDeclaredElementFromParentDeclaration([NotNull] ICSharpExpression expression)
+        {
+            expression.GetPsiServices().Locks.AssertReadAccessAllowed();
+            
+            var localVariableDeclaration =
+                LocalVariableDeclarationNavigator.GetByInitial(
+                    expression.GetContainingParenthesizedExpression()?.Parent as IVariableInitializer);
```
TessenR:
```cs
var localVariableDeclaration =
        LocalVariableDeclarationNavigator.GetByInitial(
          ExpressionInitializerNavigator.GetByValue(expression.GetContainingParenthesizedExpression()))
```

resharper/resharper-unity/src/CSharp/Feature/Services/QuickFixes/MoveQuickFixes/MonoBehaviourMoveUtil.cs
```diff
@@ -0,0 +1,263 @@
...
+
+        public static string CreateBaseName([NotNull]ICSharpExpression toMove, [CanBeNull] IDeclaredElement variableDeclaration)
+        {
+            toMove.GetPsiServices().Locks.AssertReadAccessAllowed();
+            
+            var type = toMove.Type(new ResolveContext(toMove.GetPsiModule()));
+            if (type.IsUnknown)
+                type = TypeFactory.CreateTypeByCLRName("System.Object", toMove.GetPsiModule());
+            
+            string baseName =  type.GetPresentableName(CSharpLanguage.Instance);
```
TessenR: `.GetPresentableName()` returns a string which should represent a human-readable type name in hints/warnings. It might not be a valid identifier e.g. it could return `int[]` or `MyType<T>`.
Does naming service handle such `base names`? Please, add tests for such scenarios.

resharper/resharper-unity/src/CSharp/Feature/Services/QuickFixes/MoveQuickFixes/MonoBehaviourMoveUtil.cs
```diff
@@ -0,0 +1,263 @@
...
+                        {
+                            baseName = (argument.ConstantValue.Value as string).NotNull(
+                                "argument.ConstantValue.Value as string != null");
+                        }
+                    }
+                }
+            }
+
+            return baseName;
+        }
```
TessenR: It looks like this method tries to emulate the following:
```cs
INamesCollection namesCollection;
namesCollection.Add(GetNameRootFromStringArgument(), EntryOptions { Emphasis = Good ... });
namesCollection.Add(expression, EntryOptions { ... }); // this will add roots based on the expression's type and text
```

`INamesCollection` already has the capability to suggest names based on the expression's type.
It also can suggest names based on the expression's code. It's ignored here altough I'm not sure if it's intended.

resharper/resharper-unity/src/CSharp/Daemon/Stages/PerformanceCriticalCodeAnalysis/PerformanceCriticalCodeAnalysisStage.cs
krasnotsvetov: It is not possible for Unity. MonoScript should be top-level class with same name as file name.

PR #886, opened 2018-11-04
Adding a warning if a user calls `base.OnGUI` in PropertyDrawer override

vinhui:
Added a warning for the user if they call the the `PropertyDrawer.OnGUI` method (`base.OnGUI()`) which gets added automatically. If they do call the method, they will see a `No GUI Implemented` text in the inspector which will be confusing (https://github.com/Unity-Technologies/UnityCsReference/blob/master/Editor/Mono/ScriptAttributeGUI/PropertyDrawer.cs#L35).


resharper/resharper-unity/src/CSharp/Daemon/Stages/Analysis/PropertyDrawerOnGUIBaseProblemAnalyzer.cs
```diff
@@ -0,0 +1,85 @@
...
+            var reference = expression.Reference;
+            if (reference == null)
+                return false;
+
+            var info = reference.Resolve();
+
+            if (info.ResolveErrorType != ResolveErrorType.OK)
+                return false;
+
+            var method = (info.DeclaredElement as IMethod).NotNull("info.DeclaredElement as IMethod != null");
```
citizenmatt: You can't use `NotNull` here. This is an assert and will throw if the expression is invoking something else (maybe a property or indexer, etc).

resharper/resharper-unity/src/CSharp/Daemon/Stages/Analysis/PropertyDrawerOnGUIBaseProblemAnalyzer.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            var doesNameMatch = method.ShortName == "OnGUI";
+            if (!doesNameMatch)
+                return false;
+
+            ICompiledEntity methodEntity = info.DeclaredElement as ICompiledEntity;
+            if (!(methodEntity?.Parent is CompiledTypeElement parentClass))
+                return false;
```
citizenmatt: Replace this with:

```
method.GetContainingType()?.GetClrTypeName() == KnownTypes.PropertyDrawer
```

citizenmatt: And then check for `base.` with `expression.ConditionalQualifier as IBaseExpression` and make sure it's not null.

resharper/resharper-unity/src/CSharp/Daemon/Stages/Analysis/PropertyDrawerOnGUIBaseProblemAnalyzer.cs
```diff
@@ -0,0 +1,85 @@
...
+            : base(unityApi)
+        {
+        }
+
+        protected override void Analyze(IInvocationExpression expression, ElementProblemAnalyzerData data,
+            IHighlightingConsumer consumer)
+        {
+            if (IsOnGUIBaseCall(expression) &&
+                IsInsidePropertyDrawer(expression) &&
+                IsInsideOnGUI(expression))
```
citizenmatt: Why only check inside OnGUI? Isn't every call to `PropertyDrawer.OnGUI` bad?

We could also probably get rid of the check for base type deriving from `PropertyDrawer` too, because we're checking for the `base.` qualifier, and checking that `OnGUI` is derived on `PropertyDrawer`

resharper/resharper-unity/test/data/CSharp/Daemon/Stages/Analysis/PropertyDrawerOnGUIBaseCallWarning.cs.gold
```diff
@@ -0,0 +1,32 @@
...
+        
+        OnGUI();
+        OnGUI(position, property, label);
+    }
+    
+    public void OnGUI()
+    {
+        var a = new SerializedObject(new Object());
+        var b = a.FindProperty("test");
+        base.OnGUI(new Rect(), b, new GUIContent("test"));
```
citizenmatt: I've forgotten the reason why this isn't highlighted. Might be worth adding a comment here or `PropertyDrawerOnGUIBaseProblemAnalyzer.IsInsideOnGUI` to say why.

resharper/resharper-unity/test/data/CSharp/Daemon/Stages/Analysis/PropertyDrawerOnGUIBaseCallWarning.cs.gold
```diff
@@ -0,0 +1,32 @@
...
+    }
+}
+
+public class PlainClass
+{
+    public void OnGUI(Rect position, SerializedProperty property, GUIContent label)
+    {
+        OnGUI(position, property, label);
```
citizenmatt: I think this will just be a recursive call. How about adding a base class, so the test does call `base.OnGUI`, but it's not `PropertyDrawer.OnGUI`, so doesn't highlight?

citizenmatt:
Thanks! 🎉

PR #902, opened 2018-11-15
Fix isUnity Notification

van800:
**Currently**
When you get pong-tdd from vcs it is opened as a folder.
And ApplicationPath is not populated
So Start Unity action is not availbale and Notification is saying just close project and open it via Unity.

**With this fix**
When project is opened as folder and IsUnityProjectFolder is thue, then ApplicationPath is found and passed to frontend, Start Unity is working and Notification says click here and Unity will be opened and will open Rider with solution.

### DIFF
--- a/resharper/resharper-unity/src/UnityInstallationFinder.cs
+++ b/resharper/resharper-unity/src/UnityInstallationFinder.cs
@@ -23,17 +23,17 @@ namespace JetBrains.ReSharper.Plugins.Unity
         {
             var possible = GetPossibleInstallationInfos();
             
-            var bestChoice = possible.LastOrDefault(a =>
+            var bestChoice = possible.Where(a =>
                 a.Version.Major == version.Major && a.Version.Minor == version.Minor &&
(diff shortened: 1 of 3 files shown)


resharper/resharper-unity/src/UnityInstallationFinder.cs
```diff
@@ -23,16 +23,16 @@ public FileSystemPath GetApplicationPath(Version version)
         {
             var possible = GetPossibleInstallationInfos();
             
-            var bestChoice = possible.FirstOrDefault(a =>
+            var bestChoice = possible.LastOrDefault(a =>
```
citizenmatt: I don't think `GetPossibleInstallationInfos` has a deterministic order. It looks like the directories are ordered by full path, rather than version, and then when the install info is created, it's not sorted again.

van800: Ah, it may be a small problem, I will resort then by version again.

PR #903, opened 2018-11-15
Updates to YAML support

citizenmatt:
* Support references in all asset types, not just scenes
* Group by type for event handler references in Find Usages - "Unity event handler"
* Group by type for component usage in assets - "Unity component usage"
* Group results by component and parent game objects
* Lexer based syntax highlighter for Unity YAML types (`.asset`, `.unity`, `.prefab` and `.meta` only)
* Add references for `MonoBehaviour` usage in Unity YAML types (e.g. using a script as a component)
* YAML parsing is now on by default
* Asset files only parsed if text serialisation is enabled for the project
* Disable rename refactoring for UnityEvent handlers. We'll implement it properly later (#906)

### DIFF
--- a/resharper/resharper-unity.sln.DotSettings
+++ b/resharper/resharper-unity.sln.DotSettings
@@ -43,6 +43,7 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Playmode/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=resharper/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Scriptable/@EntryIndexedValue">True</s:Boolean>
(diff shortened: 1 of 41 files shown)


resharper/resharper-unity.sln.DotSettings
```diff
@@ -43,6 +43,7 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Playmode/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=resharper/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Scriptable/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=Serisalisation/@EntryIndexedValue">True</s:Boolean>
```
krasnotsvetov: Is it correct?

citizenmatt: It's the UK English spelling :) I guess it's not important that it's in there. I can remove.

PR #913, opened 2018-11-19
mark Unity projects as projects with Android support

ArsenyChernyaev:


ArsenyChernyaev:
I'd like to add this commit to eap6

Issue #914, opened 2018-11-19
YAML parsing should respect ignored files and folders


SugoiDev:
Currently, it seems the new yaml parsing feature will not consider files/folders set in the "Elements to skip" in Editor -> Inspection Settings. 

This is undesirable when using third party assets or "pure data" assets in general.
In my particular case, I have many thousands of prefabs that contain only data (many from third parties) that should never be parsed and I have DotSettings files to ignore all third party assets in general.

Rider should not parse them if they are defined to be skipped, since this takes considerable time and will increase the time it takes to perform other operations.

Note: I'm currently on the latest nightly.

citizenmatt:
The "Elements to skip" setting is only about code analysis - it lists the items that should be skipped during code inspections. This doesn't affect any YAML files, because we not running any inspections on them. You can read more in the last paragraph of [this page](https://www.jetbrains.com/help/resharper/Reference__Options__Code_Inspection__Settings.html).

But we still index all files, even if they're marked "to skip". This is so that refactorings (e.g. renaming a symbol) still work, and don't produce broken code.

The big problem right now is that the YAML support is very naive and unoptimised, Unity projects use many large YAML files, and we're seeing long solution load times, slow find usages and very high memory usage. #921 is trying to fix these problems, and with a bit of luck, this whole issue will be obsolete. There are two main problems that need fixing:

Firstly, memory usage. We're currently building a full syntax tree for the entire contents of every YAML file, and these are getting very big, and using a lot of memory. Instead, we should be building a "sparse" tree that effectively gives you an outline and parses/creates sub-trees on demand. This should greatly reduce the amount of memory being used. #921 has already done some related work to reduce substring allocations, which has a reasonable impact on larger projects.

Secondly, when we look at the contents of the YAML files, we're looking for "references" - nodes in the tree that can resolve to a method or class (e.g. `UnityEvent` target handler methods, or `MonoScript` class references). Right now, we walk the entire syntax tree, and look at every single node, and these are very large files! This is most visible in solution load and find usages. We can be a lot smarter about this, and only look at the sections of the file that possibly contain references we're interested. #921 has already greatly improved solution load time (a test project went from 100 seconds wall clock time to 6ms for that section of solution load!). It needs a similar fix for find usages, and the "sparse" trees will also help here, too.

All of which means that if everything goes according to plan, this issue will be obsolete. Indexing of YAML files will be as quick as the other files in the solution, and you won't notice extra time in solution load.

[...]

PR #921, opened 2018-11-23
YAML performance fixes

citizenmatt:
Addresses major performance issues with large numbers of large YAML files.

Memory usage:
- [x] Use text buffers instead of allocating strings for each node in the PSI tree
- [x] Use chameleon nodes to skip creating PSI sub-trees for documents that don't need to be examined

Solution load performance:
- [x] Don't walk entire PSI tree looking for references in `UnityEvent` handler cache

Find usages performance:
- [x] Don't walk entire PSI tree looking for references in standard YAML reference searcher


SugoiDev:
Could you consider adding #914 as part of this change?

citizenmatt:
Short answer is no. I’ll reply over there with a longer answer when I’m not on my phone :)

citizenmatt:
And it’s not as bad as it sounds!

SugoiDev:
No problem, take your time!

PR #931, opened 2018-11-26
Notify user, when compilation before tests failed

van800:
Show Notification and focus UnityLogView when compilation before tests was not successful.

### DIFF
--- a/resharper/resharper-unity/src/Rider/UnitTesting/RunViaUnityEditorStrategy.cs
+++ b/resharper/resharper-unity/src/Rider/UnitTesting/RunViaUnityEditorStrategy.cs
@@ -21,6 +21,7 @@ using JetBrains.ReSharper.UnitTestFramework.Strategy;
 using JetBrains.ReSharper.UnitTestProvider.nUnit.v30;
 using JetBrains.ReSharper.UnitTestProvider.nUnit.v30.Elements;
 using JetBrains.Rider.Model;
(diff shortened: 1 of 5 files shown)


resharper/resharper-unity/src/Rider/UnitTesting/RunViaUnityEditorStrategy.cs
```diff
@@ -133,6 +145,10 @@ Task IUnitTestRunStrategy.Run(IUnitTestRun run)
                         if (!result.Result)
                         {
                             tcs.SetException(new Exception("There are errors during compilation in Unity."));
+                            
+                            var notification = new RdNotificationEntry("Compilation failed", "Scripts compilation in Unity failed, so tests were not started.", true, RdNotificationEntryType.INFO);
+                            mySolution.Locks.ExecuteOrQueueEx(run.Lifetime, "RunViaUnityEditorStrategy compilation failed", () => myNotificationsModel.Notification.Fire(notification));
+                            myUnityHost.PerformModelAction(model => model.ActivateUnityLogView.Fire());
```
citizenmatt: Minor nitpick: "Script compilation" instead of "scripts".

And move `notification` inside the lambda passed to `ExecuteOrQueueEx`. It's a ridiculously tiny optimisation, and won't make any noticeable difference, but I've got to comment on something :)

Issue #922, opened 2018-11-23
Method marked as never used when used by child classes


Nezz:
I just tried the method usages feature in EAP 6 and it is a great improvement. I have found an issue though. If method used by buttons are declared in an abstract MonoBehaviour, they get marked as never used:
<img width="566" alt="screen shot 2018-11-23 at 15 37 17" src="https://user-images.githubusercontent.com/431167/48958495-c0b20f00-ef35-11e8-9e49-9ac426576549.png">
Find Usages correctly discovers the actual usages though, which are child classes inheriting from the abstract base class.

PR #939, opened 2018-11-29
Updates to cache handling and property event handlers

citizenmatt:
* Fix exception on startup - [RIDER-21929](https://youtrack.jetbrains.com/issue/RIDER-21929) (mistyped the ticket in the commit)
* Mark event handlers as in use if they're declared on a base type - #922 
* Fix highlighting for property used as event handler - gets gutter icon/code insight marker
* Fix find usages for properties used as event handlers

Plus sneaky, unrelated fix to stop the Generate event function dialog selecting all items by default when called from the gutter icon or code code insight marker. See [RIDER-22211](https://youtrack.jetbrains.com/issue/RIDER-22211)

### DIFF
--- a/rider/.idea/encodings.xml
+++ b/rider/.idea/encodings.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="Encoding" addBOMForNewFiles="with NO BOM" />
+</project>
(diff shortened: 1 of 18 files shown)


rider/.idea/encodings.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
krasnotsvetov: Do we really need this file?

citizenmatt: Possibly not. But I was fed up with it in my "unversioned files". Looking at the file, and the options page that seems to drive it, it looks like it's something that should be shared. And since IntelliJ has no sensible scheme for knowing if a file should be committed or not, I flipped a coin and it ended in.

PR #942, opened 2018-11-30
fix RIDER-22276 + Show Unity version in Start Unity action

van800:
…etBrains.ReSharper.Plugins.Unity.UnityApi. Exception has been thrown by the target of an invocation. Could not find any recognizable digits.

### DIFF
--- a/resharper/resharper-unity/src/UnityVersion.cs
+++ b/resharper/resharper-unity/src/UnityVersion.cs
@@ -11,6 +11,7 @@ using JetBrains.ProjectModel.Properties;
 using JetBrains.ProjectModel.Properties.Managed;
 using JetBrains.ReSharper.Plugins.Unity.ProjectModel.Caches;
 using JetBrains.Util;
+using JetBrains.Util.Logging;
 
 namespace JetBrains.ReSharper.Plugins.Unity
 {
@@ -28,7 +29,7 @@ namespace JetBrains.ReSharper.Plugins.Unity
         {
             myUnityProjectFileCache = unityProjectFileCache;
             mySolution = solution;
-            
+
             if (locks.Dispatcher.IsAsyncBehaviorProhibited) // for tests
                 return;
 
@@ -132,7 +133,16 @@ namespace JetBrains.ReSharper.Plugins.Unity
             Version version = null;
             if (match.Success)
             {
-                var type = Convert.ToInt32(groups["type"].Value+groups["revision"], 16);
+                var type = 0;
+                try
+                {
+                    type = Convert.ToInt32(groups["type"].Value + groups["revision"].Value, 16);
+                }
+                catch (Exception e)
+                {
+                    Logger.GetLogger<UnityVersion>().Error($"Unable to parse part of version. type={groups["type"].Value} revision={groups["revision"].Value}", e);
+                }
+
                 version = Version.Parse($"{groups["major"].Value}.{groups["minor"].Value}.{groups["build"].Value}.{type}");
             }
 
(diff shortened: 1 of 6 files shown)


resharper/resharper-unity/src/UnityVersion.cs
```diff
@@ -132,7 +133,16 @@ public static Version Parse(string input)
             Version version = null;
             if (match.Success)
             {
-                var type = Convert.ToInt32(groups["type"].Value+groups["revision"], 16);
+                var type = 0;
+                try
+                {
+                    type = Convert.ToInt32(groups["type"].Value + groups["revision"].Value, 16);
```
citizenmatt: What is this trying to do? Convert something like "f2" into decimal as though it's a hexadecimal string? This will fail with anything that isn't `a` - `f`, and I know Unity have used `p` for patch versions. 

Do we need to know these revision number/types? If so, we'll need to come up with an encoding scheme, e.g. convert the type character into an ASCII byte and store it in the top half of an int, with the revision in the lower half.

van800: fixed

resharper/resharper-unity/test/src/CSharp/UnityVersionTest.cs
```diff
@@ -0,0 +1,40 @@
...
+            var marketingVersion = "2018.2.13";
+            var realVersion = Version.Parse("2018.2.13.0");
+            
+            Assert.AreEqual(marketingVersion, Unity.UnityVersion.VersionToString(realVersion));
+        }
+
+        [Test]
+        public void SortTest()
+        {
+            var mVersions = new[] {"2018.2.13p1", "2018.2.13a20", "2018.2.13f20", "2018.2.13b20", "2018.2.13p20"};
```
citizenmatt: Add in some extra values with different ranges, e.g. `f1` and `b10` that would break with the existing rules

resharper/resharper-unity/src/UnityVersion.cs
```diff
@@ -132,12 +134,50 @@ public static Version Parse(string input)
             Version version = null;
             if (match.Success)
             {
-                var type = Convert.ToInt32(groups["type"].Value+groups["revision"], 16);
-                version = Version.Parse($"{groups["major"].Value}.{groups["minor"].Value}.{groups["build"].Value}.{type}");
+                var typeWithRevision = "0";
+                try
+                {
+                    var typeChar = groups["type"].Value.ToCharArray()[0];
+                    var shiftedChar = 16 + typeChar; // Because `f1` = `1021` and `b10` = `9810`, which will break sorting
```
citizenmatt: This needs to be `typeChar << 16`

resharper/resharper-unity/src/UnityVersion.cs
```diff
@@ -132,12 +134,50 @@ public static Version Parse(string input)
...
+        public static string VersionToString(Version version)
+        {
+            var type = string.Empty;
+            var rev = string.Empty;
+            try
+            {
+                var revisionString = version.Revision.ToString(); // first 3 is char, next 1+ ones - revision
+                if (revisionString.Length > 3)
+                {
+                    var charValue = Convert.ToInt32(revisionString.Substring(0, 3)) - 16;
```
citizenmatt: You can't use 3 characters here. `b` is 98. You'll need to do the opposite of `<< 16`:

```
char charValue = (char)(version.Revision >> 16) & 0xFF;
```

resharper/resharper-unity/src/UnityVersion.cs
```diff
@@ -132,12 +134,50 @@ public static Version Parse(string input)
...
+                    type = ((char)charValue).ToString();
+                    rev = Convert.ToInt32(revisionString.Substring(3)).ToString();
```
citizenmatt: And you can't use 3 characters here. Just use `version.Revision & 0xFFFF`

PR #946, opened 2018-12-05
Fixes

van800:
fix RIDER-22361
avoid ConnectionTracker and UnityRefreshTracker doing things outside Unity

### DIFF
--- a/resharper/resharper-unity/src/Rider/UnityRefreshTracker.cs
+++ b/resharper/resharper-unity/src/Rider/UnityRefreshTracker.cs
@@ -107,24 +107,31 @@ namespace JetBrains.ReSharper.Plugins.Unity.Rider
             if (solution.GetData(ProjectModelExtensions.ProtocolSolutionKey) == null)
                 return;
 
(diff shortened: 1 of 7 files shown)


resharper/resharper-unity/src/Rider/UnityRefreshTracker.cs
```diff
@@ -107,24 +107,31 @@ public class UnityRefreshTracker
             if (solution.GetData(ProjectModelExtensions.ProtocolSolutionKey) == null)
                 return;
 
-            unitySolutionTracker.IsUnityProject.ViewNotNull(lifetime, (lf, args) =>
+            // Rgc.Guarded - beware RIDER-15577
+            myGroupingEvent = solution.Locks.GroupingEvents.CreateEvent(lifetime, "UnityRefresherOnSaveEvent",
```
krasnotsvetov: Should we add this under `unitySolutionTracker.IsUnityProjectFolder.AdviseOnce`?
Why do not put all constructor logic under `unitySolutionTracker.IsUnityProjectFolder.AdviseOnce`?

PR #950, opened 2018-12-06
Bug fixes

krasnotsvetov:
1. fix #RIDER-22196
2. fix #RIDER-22114
3. fix #RIDER-21924
4. fix RIDER-22342

Add tooltip when unity is not running and user click "Show usages in Unity"
Add "Start Unity" action in lens' context menu 
Update algorithm which checks possibility to move expression outside method/loop


resharper/resharper-unity/src/Rider/CodeInsights/RiderUnityImplicitUsageHighlightingContributor.cs
```diff
@@ -97,27 +115,39 @@ public override IEnumerable<BulbMenuItem> CreateAdditionalMenuItem(IDeclaration
...
             {
                 myDeclaredElement = method;
                 myCreator = creator;
+                myTooltipManager = tooltipManager;
+                myTracker = tracker;
             }
             
             protected override Action<ITextControl> ExecutePsiTransaction(ISolution solution, IProgressIndicator progress)
             {
+                if (myTracker.LastCheckResult == UnityEditorState.Disconnected)
```
van800: Maybe not accurate. As I remember some other states are also disconnected, consider the following
if (result.Result != UnityEditorState.Refresh && result.Result != UnityEditorState.Disconnected)

PR #954, opened 2018-12-08
Track Unity and UnityClassLib project types

van800:
I can't really test it right now, because of failed compilation


resharper/resharper-unity/src/Rider/UnityProjectTechnologyProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+        
+        public IEnumerable<string> GetProjectTechnology(IProject project)
+        {
+            if (myUnitySolutionTracker.IsUnityGeneratedProject.Value) yield return "Unity";      
+        }
+    }
+    
+    [ShellComponent]
+    public class UnityClassLibProjectTechnologyProvider : IProjectTechnologyProvider
```
xtmq: Why 2 different providers?

van800: please see the new version

PR #959, opened 2018-12-12
Fit & finish

citizenmatt:
Final polish:

- [x] Update icons
- [x] Fix up wording


resharper/resharper-unity/src/Rider/CodeInsights/RiderUnityImplicitUsageHighlightingContributor.cs
```diff
@@ -87,12 +87,12 @@ private void AddHighlighting(IHighlightingConsumer consumer, AbstractUnityCodeIn
             var extraActions = new List<CodeLensEntryExtraActionModel>();
             if (!myConnectionTracker.IsConnectionEstablished())
             {
-                extraActions.Add(new CodeLensEntryExtraActionModel("Unity is off", null));
+                extraActions.Add(new CodeLensEntryExtraActionModel("Unity is not running", null));
```
krasnotsvetov: No! 
Use shorter string, because there is no space for long strings in context action for code vision

Issue #581, opened 2018-06-05
Review editor-plugin installation rules for non-unity-generated solution


van800:
More info here https://youtrack.jetbrains.com/issue/RIDER-16676#comment=27-2902532

As I see it, we need the minimized Unity UI only for unity-generated solution.
But all other features can be beneficial if ProtocolInstance.json is in the expected location.

- [x] Plugin is installed for any solution with reference to UnityEngine and existing Assets folder.
Related usability issue https://youtrack.jetbrains.com/issue/RIDER-16757
- [x] Allow Unity Editor to connect to multiple Rider backend instances.
- [x] Provide a manual install/update button for when the user has unchecked the automatic update setting.
- [x] Install plugin without waiting for Solution Load
- [x] Handle outdated protocol and the automatic update is unchecked. (use Protocol.OutOfSyncModels)
 - [x] After new Rider installation EditorPlugin is updated also, but Refresh is required for it to be loaded in Unity.

PR #963, opened 2018-12-13
Handle outdated protocol and the automatic update is unchecked.

van800:
Fix for https://github.com/JetBrains/resharper-unity/issues/581

### DIFF
--- a/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/OutOfSyncEditorNotification.kt
+++ b/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/OutOfSyncEditorNotification.kt
@@ -17,9 +17,9 @@ class OutOfSyncEditorNotification(project: Project, unityHost: UnityHost): Lifet
 
     init {
         unityHost.model.onEditorModelOutOfSync.adviseNotNull(componentLifetime) {
-            val message = "UnityEditor connection is out of sync, but automatic plugin update is disabled."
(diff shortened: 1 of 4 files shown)


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/OutOfSyncEditorNotification.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.jetbrains.rider.util.reactive.adviseNotNull
+
+class OutOfSyncEditorNotification(project: Project, unityHost: UnityHost): LifetimedProjectComponent(project) {
+    companion object {
+        private val notificationGroupId = NotificationGroup.balloonGroup("Unity connection is out of sync")
+    }
+
+    init {
+        unityHost.model.onEditorModelOutOfSync.adviseNotNull(componentLifetime) {
+            val message = "UnityEditor connection is out of sync, but automatic plugin update is disabled."
```
citizenmatt: How about?

> **Unity editor plugin update required**
>
> The Unity editor plugin is out of date and automatic plugin updates are disabled. Advanced Unity integration features are unavailable until the plugin is updated.

van800: Even though it doesn't reveal all the technical details, but it makes more sense for the user. Thanks!

Issue #383, opened 2018-02-28
Please add drag-n-drop file in Unity Explorer


van800:

van800:
https://youtrack.jetbrains.com/issue/RIDER-15697

sdarkwell:
I agree with @Humberd. I have stopped using the Unity Explorer, because when I try to use it, I find myself constantly switching back to the Solution Explorer. It becomes more cumbersome than helpful.

van800:
On 2018.2EAP1 drag-n-drop is working only partially.
I can't move file to Editor folder here, because there are no other cs files in it, thus no Assembly-Editor.csproj

![2018-06-29_12-30-41](https://user-images.githubusercontent.com/1482681/42088029-63c716fc-7b98-11e8-9712-1b174ce74836.gif)

vinhui:
It also seems it won't allow me to move a file to a `asset`/`assets` sub folder. For example `Assets/Scripts/MyModule/Assets`.

van800:
just for internal reference https://github.cds.internal.unity3d.com/unity/com.unity.ide.rider/pull/34
fix in Rider package should help with drag-drop into empty folders

van800:
Drag-n-drop for folders, which are out of the project was fixed in Rider 2019.3. Would be available in first EAP.

sebas77:
It seems I am having this issue with 2019.2.3 I have to open the solution explorer to be able to move an existing file into a new folder. This also mean that the perforce plugin won't be able to do its job. This is happening also with the file explorer view, I don't think it's unity related.

van800:
I stated `Rider 2019.3`.

sebas77:
yes but the title also states that it happens only in the Unity Explorer, so I added that is not my case in case is a useful add :)

van800:
If you have an issue in non-unity project, please report it in https://youtrack.jetbrains.com

PR #971, opened 2018-12-20
[WIP] Add empty folders to first csproj

van800:
Users are complaining about refactorings and drag-n-drop not working for empty folders. The reason is that empty folders are not in any csproj, so nothings works.
This PR adds all folders without cs files to **first** csproj. (I decided to avoid complicated logic for adding folders to corresponding csproj-s, specifically because csproj can be missing, but folder can be there and user may want to drag-n-drop file to it.)

https://youtrack.jetbrains.com/issue/RIDER-23169
https://github.com/JetBrains/resharper-unity/issues/383 (edited)


van800:
We solved drag-n-drop in a different way in 2019.3

PR #973, opened 2018-12-24
Heuristic, which disables yaml parsing, if unity content is large

krasnotsvetov:


resharper/resharper-unity/src/Settings/UnitySettings.cs
```diff
@@ -26,5 +26,13 @@ public class UnitySettings
...
         [SettingsEntry(GutterIconMode.CodeInsightDisabled, "Unity highlighter scheme for editor.")]
         public GutterIconMode GutterIconMode;
+
+        
+        // TODO : merge this two options : any converter for settings?
+        [SettingsEntry(true, "Enables heuristic which can disable yaml for performance reason")]
+        public bool EnableYamlHeuristic;
+        
+        [SettingsEntry(true, "Enables syntax error highlighting, brace matching and more of YAML files for Unity")]
+        public bool EnableYamlParsing;
```
citizenmatt: Do we need both of these? `EnableYamlParsing` sounds just like the one in `YamlSupport`. I thought we only needed another setting with a three state - enabled (default), heuristically disabled and force enabled.

resharper/resharper-unity/src/Yaml/Psi/Modules/UnityExternalFilesModuleProcessor.cs
```diff
@@ -83,11 +90,8 @@ public void OnUnityProjectAdded(Lifetime projectLifetime, IProject project)
             // Note that this means we process meta and asset files for class library projects, which likely won't have
             // asset files. We can't use IsUnityGeneratedProject because any project based in the Packages folder or
             // using 'file:' won't be processed.
-            if (!myAssetSerializationMode.IsForceText || !myYamlSupport.IsParsingEnabled.Value ||
-                !project.IsUnityProject())
-            {
+            if (!project.IsUnityProject())
```
citizenmatt: We shouldn't walk the file system if asset serialisation mode is not set to "ForceText"

resharper/resharper-unity/src/Yaml/Psi/Modules/UnityYamlDisableStrategy.cs
```diff
@@ -0,0 +1,103 @@
...
+                if (!IsYamlParsingAvailable())
+                {
+                    YamlParsingEnabled.Value = false;
+                } 
+                else if (IsAnyFilePreventYamlParsing(solutionDirectory) || myTotalSize > YamlFileTotalSizeThreshold)
+                {
+                    YamlParsingEnabled.Value = false;
+                    CreateNotification();
+                }
+                myEnabled.Value = false;
```
citizenmatt: Does this mean we only ever run the heuristic once per solution? What if a file is added that is e.g. too large? If it is intended to be a run once setting, the name in the setting file should reflect that - `EnableYamlHeuristic` doesn't indicate that it's only run once.

resharper/resharper-unity/src/Yaml/UnityYamlEnabled.cs
```diff
@@ -0,0 +1,20 @@
...
+    public class UnityYamlEnabled
+    {
+        public readonly IProperty<bool> YamlParsingEnabled;
+        public UnityYamlEnabled(Lifetime lifetime, ISolution solution, ISettingsStore settingsStore)
+        {
+            var settings = settingsStore.BindToContextLive(lifetime,
+                ContextRange.ManuallyRestrictWritesToOneContext(solution.ToDataContext()));
+            YamlParsingEnabled = settings.GetValueProperty(lifetime, (UnitySettings key) => key.EnableYamlParsing);
+        }
+    }
```
citizenmatt: This should probably do something to override/update `YamlSupport.EnableYamlParsing` - if that's been set to false in 2018.3.0, then it can't be set to true again in this update.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/YamlHugeFileNotification.kt
```diff
@@ -0,0 +1,55 @@
...
+import com.jetbrains.rider.model.ScriptCompilationDuringPlay
+import com.jetbrains.rider.plugins.unity.UnityHost
+import com.jetbrains.rider.util.reactive.adviseNotNullOnce
+import com.jetbrains.rider.util.reactive.fire
+import javax.swing.event.HyperlinkEvent
+
+class YamlHugeFileNotification(project: Project, private val unityHost: UnityHost): LifetimedProjectComponent(project) {
+
+    companion object {
+        private val notificationGroupId = NotificationGroup.balloonGroup("Unity Enable Yaml")
```
citizenmatt: Does this string get shown to the user?

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/YamlHugeFileNotification.kt
```diff
@@ -0,0 +1,55 @@
...
+        }
+    }
+
+    private fun showNotificationIfNeeded(){
+
+        val message = """Due to the size of the project, parsing of Unity scenes, assets and prefabs has been disabled. Usages of C# code in these files will not be detected. Re-enabling can impact initial file processing.
+            <ul style="margin-left:10px">
+              <li><a href="turnOnYamlParsing">Turn on anyway</a></li>
+            </ul>
+            <a href="doNotShow">Do not show</a> this notification for this solution.
```
citizenmatt: No need to "do not show". Probably worth making "turn on anyway" as an action link rather than a link inside the text.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/YamlHugeFileNotification.kt
```diff
@@ -0,0 +1,55 @@
...
+            """
+
+        val generalSettings = GeneralSettings.getInstance()
+        if (generalSettings.isAutoSaveIfInactive || generalSettings.isSaveOnFrameDeactivation){
```
citizenmatt: Auto save?

resharper/resharper-unity/src/Settings/UnitySettings.cs
krasnotsvetov: I wanted to merge it, but don't find a good solution. 
There is problem with checkbox, this settings is binded to it.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/YamlHugeFileNotification.kt
krasnotsvetov: oops.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/YamlHugeFileNotification.kt
krasnotsvetov: Yes, forgot to remove it. Thanks!

resharper/resharper-unity/src/Yaml/UnityYamlEnabled.cs
krasnotsvetov: I hope, that we can drop migration here. We will disable yaml parsing in most part of cases where performance can be affected via heuristic

resharper/resharper-unity/src/Yaml/Psi/Modules/UnityYamlDisableStrategy.cs
krasnotsvetov: Yes.
We discussed it today with @korifey and decided that if user adds very large file, he understands that Rider processing this file. 
Also, it will be very strange when yaml features are enabled and after some time (for user, he doesn't know our idea with disabling) yaml features are disabled

resharper/resharper-unity/src/Yaml/Psi/Modules/UnityExternalFilesModuleProcessor.cs
krasnotsvetov: This allows us to handle change of serialisation mode in Unity without solution reopenning.

Why do you think, that adding psiSourceFiles can affect performance?

resharper/resharper-unity/src/Yaml/Psi/Modules/UnityExternalFileProperties.cs
```diff
@@ -1,15 +1,23 @@
...
 
 namespace JetBrains.ReSharper.Plugins.Unity.Yaml.Psi.Modules
 {
     public class UnityExternalFileProperties : IPsiSourceFileProperties
     {
+        private readonly IProperty<bool> myEnabled;
+
+        public UnityExternalFileProperties(IProperty<bool> enabled)
+        {
+            myEnabled = enabled;
```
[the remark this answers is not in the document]
krasnotsvetov: Yes,
1. When `enabled` returns false, build one PsiSourceFile will not parse document, so caches will drop data
2. When `enabled` returns true, build one PsiSourceFile will return actual tree, caches will be builded from correct data without solution restarting

resharper/resharper-unity/src/Yaml/Psi/Caches/UnityCacheUpdater.cs
```diff
@@ -0,0 +1,50 @@
...
+                        {
+                            if (psiSourceFiles.Any())
+                            {
+                                locks.ExecuteWithWriteLock(() => changeManager.ExecuteAfterChange(() =>
+                                {
+                                    var changeBuilder = new PsiModuleChangeBuilder();
+                                    foreach (var sourceFile in psiSourceFiles)
+                                    {
+                                        if (sourceFile.IsValid())
+                                            changeBuilder.AddFileChange(sourceFile, PsiModuleChange.ChangeType.Modified);
```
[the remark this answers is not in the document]
krasnotsvetov: PsiSourceFiles should be never removed from PsiModule.
Cache should be invalidated, I will fix it.

resharper/resharper-unity/src/Settings/UnitySettings.cs
```diff
@@ -26,5 +26,13 @@ public class UnitySettings
...
         [SettingsEntry(GutterIconMode.CodeInsightDisabled, "Unity highlighter scheme for editor.")]
         public GutterIconMode GutterIconMode;
+
+        
+        // TODO : merge this two options : any converter for settings?
+        [SettingsEntry(false, "Enables heuristic which can disable yaml for performance reason")]
+        public bool IsYamlHeuristicApplied;
+        
+        [SettingsEntry(true, "Enables syntax error highlighting, brace matching and more of YAML files for Unity")]
+        public bool IyYamlParsingEnabled;
```
korifey: 'Iy' errata

resharper/resharper-unity/src/Yaml/Psi/Modules/UnityExternalFilesModuleProcessor.cs
```diff
@@ -70,6 +74,9 @@ public class UnityExternalFilesModuleProcessor : IChangeProvider, IUnityReferenc
             if (!mySolutionDirectory.IsAbsolute)
                 mySolutionDirectory = solution.SolutionDirectory.ToAbsolutePath(FileSystemUtil.GetCurrentDirectory());
 
+            unityYamlDisableStrategy.Run(mySolutionDirectory);
```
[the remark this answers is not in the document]
citizenmatt: I don't recall discussing moving it outside of `OnUnityProjectAdded`. We can easily do this as part of `OnUnityProjectAdded` - on first call, walk `Assets`, `Packages` and `ProjectSettings`, collect all data and then act on it - add the filenames to a list and create them after the walk is finished.

If we're concerned about performance in very large projects it seems really weird to do this twice, especially as part of a class constructor. It feels like it would nearly double the "Preparing Unity project" stage (it's doing different things with the data, and hopefully the file system cache would improve things, but it's still walking the solution structure twice).

resharper/resharper-unity/src/Yaml/Psi/Modules/UnityYamlDisableStrategy.cs
citizenmatt: I don't think that's the right approach - if someone else in the team adds a large scene, or in 6 months time a scene has grown in complexity, we'll ignore it. I think if we're going to have this heuristic, we should run it every time. I think it can easily be a part of the initial call to `OnUnityProjectAdded`

PR #987, opened 2019-01-09
Add extra logging for play mode and refresh

citizenmatt:
Add a little bit of logging around switching between play/pause mode, as well as during refresh.

### DIFF
--- a/unity/EditorPlugin/PluginEntryPoint.cs
+++ b/unity/EditorPlugin/PluginEntryPoint.cs
@@ -35,8 +35,8 @@ namespace JetBrains.Rider.Unity.Editor
     static PluginEntryPoint()
     {
       PluginSettings.InitLog(); // init log before doing any logging
-      ourLogEventCollector = new UnityEventCollector(); // start collecting Unity messages asap 
-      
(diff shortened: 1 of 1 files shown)


unity/EditorPlugin/PluginEntryPoint.cs
```diff
@@ -337,60 +353,63 @@ private static void AdviseRefresh(EditorPluginModel model)
...
             model.Pause.SetValue(isPaused);
+          }
         });
       });
-      isPlayingAction(); // get Unity state
-      model.Play.AdviseNotNull(connectionLifetime, play =>
+
+      syncPlayState();
+
+      model.Play.Advise(connectionLifetime, play =>
```
van800: What is the reason for changing AdviseNotNull to Advise here?

van800: Thanks for clarification.

korifey: Approved. Anyway, I guess we should add `where` clause to this method to suppress boxing.

PR #1001, opened 2019-01-18
Fix selection of property name for Unity 2018

krasnotsvetov:


resharper/resharper-unity/src/Yaml/Psi/UnitySceneProcessor.cs
```diff
@@ -125,18 +125,28 @@ public static string GetUnityObjectTag(IYamlDocument document)
...
         {
-            return myVersion.GetActualVersionForSolution().Major == 2017
-                ? UnityYamlConstants.CorrespondingSourceObjectProperty2017
-                : UnityYamlConstants.CorrespondingSourceObjectProperty;
+            if (myVersion.GetActualVersionForSolution().Major == 2017)
+            {
+                return document.GetUnityObjectPropertyValue(UnityYamlConstants.CorrespondingSourceObjectProperty2017)?.AsFileID();
+            } else
+            {
+                return document.GetUnityObjectPropertyValue(UnityYamlConstants.CorrespondingSourceObjectProperty2017)?.AsFileID() ??
```
citizenmatt: How about just having the `return` statement, no `if`? Worth extracting this pattern into a method? Perhaps look for the 2018 version first?

krasnotsvetov: Micro-optimzation for 2017 version, it should not containt properties names from 2018

krasnotsvetov: Also, I reverted checking for 2018 version

Issue #820, opened 2018-10-02
Show notification if there is no player log to open


citizenmatt:
There isn't always a player log file available to open. When selecting the item from the Unity tool window options cog icon, if the file doesn't exist, we should notify the user, also showing the path we tried to open.

citizenmatt:
See #1006

PR #1006, opened 2019-01-23
#820 Add notification to inform user there is no player log to open.

ajon542:


van800:
Thank you for the PR, could you please sign the https://www.jetbrains.com/agreements/cla/? Thanks!

ajon542:
No problems, I signed it about 2 hours ago. Is there a reference number for the completed DocuSign document that I can send or does it take a bit of time to go through?

van800:
Could you please polish it a bit?
1. Format message a bit, like:  `"$f" is not present.`
2. Handle the case when `project.solution.rdUnityModel.playerLogPath.valueOrNull` is null.
3. RiderUnityOpenEditorLogAction would also benefit from showing a similar notification.

ajon542:
I spent a bit of time trying to understand how the value could ever be null as well as getting a feel the communication between the Editor plugin, Resharper backend and the IntelliJ frontend.
`project.solution.rdUnityModel.playerLogPath.valueOrNull`

The values for player/editor log are defined as not nullable in the editor plugin and backend. If the value is set to null in the editor plugin, it will be logged to file as an error, but, will propagate through to the frontend and be null at the time the user attempts to access the log. If the value is somehow null in the backend, a BackendException will be raised and the user prompted to report the issue.

It seems in this scenario, there is not much the user can do to resolve this error themselves. I think we would want to log an error and possibly prompt the user to report the error. Does this sound reasonable?

van800:
@ajon542 That sounds reasonable.

PR #1034, opened 2019-02-13
Frontend test with Protocol connection

van800:
Test is disabled for now, because Unity license check is not yet solved on agent.
But I want to merge this, because of couple of fixes and refactorings done.


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/actions/StartUnityAction.kt
```diff
@@ -31,15 +31,21 @@ open class StartUnityAction : DumbAwareAction("Start Unity", "Start Unity with c
     }
 
     companion object {
-        fun StartUnity(project: Project, vararg args: String) {
-            val appPath = UnityInstallationFinder.getInstance(project).getApplicationPath() ?: return
+        fun StartUnity(project: Project, vararg args: String): Process? {
```
citizenmatt: Standard naming would be `startUnity`

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/actions/StartUnityAction.kt
```diff
@@ -31,15 +31,21 @@ open class StartUnityAction : DumbAwareAction("Start Unity", "Start Unity with c
...
+            val appPath = UnityInstallationFinder.getInstance(project).getApplicationPath() ?: return null
+            return StartUnity(appPath, project, args)
+        }
+
+        fun StartUnity(appPath: java.nio.file.Path, project: Project, args: Array<out String>): Process? {
```
citizenmatt: `startUnity`

rider/src/test/kotlin/ConnectionTest.kt
```diff
@@ -0,0 +1,52 @@
...
+import kotlin.test.assertNotNull
+
+class ConnectionTest : BaseTestWithSolution() {
+    override fun getSolutionDirectoryName(): String {
+        return "SimpleUnityProject"
+    }
+
+    override val waitForCaches = true;
+
+    var unityPackedUrl = "https://repo.labs.intellij.net/dotnet-rider-test-data/Unity_2018.3.4f1_stripped_v4.zip";
```
citizenmatt: I don't like that we have to depend on a repackaged Unity on a private URL. Means no-one external to the company can run this test, but I don't really know what the alternative is.

citizenmatt: Perhaps use installation finder to try and find 2018.3 locally, and if not available, download and use this?

rider/testData/solutions/SimpleUnityProject/Assembly-CSharp.csproj
```diff
@@ -0,0 +1,67 @@
...
+  <PropertyGroup>
+    <NoConfig>true</NoConfig>
+    <NoStdLib>true</NoStdLib>
+    <AddAdditionalExplicitAssemblyReferences>false</AddAdditionalExplicitAssemblyReferences>
+    <ImplicitlyExpandNETStandardFacades>false</ImplicitlyExpandNETStandardFacades>
+    <ImplicitlyExpandDesignTimeFacades>false</ImplicitlyExpandDesignTimeFacades>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="UnityEngine">
+      <HintPath>F:\Program Files\2018.3.3f1\Editor\Data\Managed/UnityEngine/UnityEngine.dll</HintPath>
```
citizenmatt: Does this `HintPath` cause issues when the path could be anything?

rider/src/test/kotlin/ConnectionTest.kt
citizenmatt: This tool might help, too: https://github.com/DragonBox/u3d

PR #1058, opened 2019-02-22
Support edge case with tests which have no difference except being in different assemblies

van800:
Fix RIDER-24636 Rider runs wrong tests if two projects has the same namespace name


unity/EditorPlugin/AfterUnity56/UnitTesting/TestEvent.cs
```diff
@@ -23,14 +23,16 @@ public TestEvent(EventType type, TestInternalEvent tEvent)
   public class TestInternalEvent
   {
     public string myID;
+    public readonly string myAssemblyName;
```
citizenmatt: Make the other field `readonly` too?

van800: Well spotted! I have to actually remove readonly here. Otherwise it will not get restored after AppDomain reload.

PR #1080, opened 2019-03-05
RIDER-24005 [Linux/Mac] Dialog with proposition to install mono for Unity project

van800:
1. Only Mac for now
2. IsUnityGeneratedProject
3. TargetFrameworkVersion >=4.7
4. mono version is less than 5.16

I wonder if it maybe requires a setting to avoid the check, like "Never show again"?

Things to check:
1. If msbuild is missing, dialog should be shown.
2. AutoSaveNotification should not be shown, if non-of-projects was loaded.
3. Sometimes it locks with empty dialog and progress.


resharper/resharper-unity/src/Rider/MonoInstallTrigger.cs
```diff
@@ -0,0 +1,77 @@
...
+                solution.Locks.Tasks.Queue(lifetime, () =>
+                {
+                    var wellKnownMonoRuntimes = MonoRuntimeDetector.DetectWellKnownMonoRuntimes();
+                    var installedValidMono = wellKnownMonoRuntimes
+                        .Any(runtime =>
+                        {
+                            var version = new Version(0, 0); // if we fail to parse version - consider it is old
+                            try
+                            {
+                                version = new Version(ProcessOutputUtil.ExtractMonoVersion(runtime.ExePath));
```
citizenmatt: We should be able to use `Version.TryParse` here.

resharper/resharper-unity/src/Rider/MonoInstallTrigger.cs
```diff
@@ -0,0 +1,77 @@
...
+                            }
+                            catch (Exception e)
+                            {
+                                logger.Warn(e);
+                            }
+
+                            return version.Major >= 5 && version.Minor >= 16;
```
citizenmatt: Can you add a comment why it has to be >= 5.16?

resharper/resharper-unity/src/Rider/MonoInstallTrigger.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+                foreach (var project in solution.GetAllProjects())
+                {
+                    var isNetFramework47 = project.TargetFrameworkIds
+                        .Any(x => x.IsNetFramework && x.Version.Major >= 4 && x.Version.Minor >= 7);
+                
+                    if (isNetFramework47)
+                    {
+                        return true;
+                    }
```
citizenmatt: Simplify to just `return project.TargetFrameworkIds.Any(…);`

Also, add a comment to why we're looking for target framework >= 4.7

citizenmatt:
Also, can you add this to the CHANGELOG.md, please?

resharper/resharper-unity/src/Rider/MonoInstallTrigger.cs
van800: done

PR #1099, opened 2019-03-24
Fix exception for file: and git based packages

### LINKED ISSUE #1095: Unity Explorer disappeared
image
Where item "Unity" ?

Version 2019.1.0.17

citizenmatt:
This PR will:

- [x] Correctly handle local `file:` and git based packages in Unity Explorer
- [x] Unresolved packages will no longer crash the Unity Explorer, but be displayed as unresolved nodes
- [x] Simplify sorting nodes in the Unity Explorer tree

Fixes #1095, fixes [RIDER-25971](https://youtrack.jetbrains.com/issue/RIDER-25971)

### DIFF
--- a/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/explorer/AssetsRoot.kt
+++ b/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/explorer/AssetsRoot.kt
@@ -1,0 +1,142 @@
+package com.jetbrains.rider.plugins.unity.explorer
+
+import com.intellij.ide.projectView.PresentationData
+import com.intellij.ide.util.treeView.AbstractTreeNode
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.text.StringUtil
+import com.intellij.openapi.vfs.VirtualFile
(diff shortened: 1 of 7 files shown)


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/explorer/AssetsRoot.kt
```diff
@@ -0,0 +1,137 @@
...
+        val visitor = object : ProjectModelNodeVisitor() {
+            override fun visitReference(node: ProjectModelNode): Result {
+                if (node.isAssemblyReference()) {
+                    val keys = referenceNames.getOrCreate(node.name) { _ -> arrayListOf() }
+                    keys.add(node.key)
+                }
+                return Result.Stop
+            }
+        }
+        visitor.visit(project!!)
```
van800: Could this ever be called when project is null?

citizenmatt: No, we only ever create the Unity explorer and nodes when we have a project.

PR #1108, opened 2019-03-27
fix RIDER-26102, RIDER-26109, RIDER-26103 

van800:


resharper/resharper-unity/src/ProjectExtensions.cs
```diff
@@ -40,7 +40,7 @@ public static bool HasUnityReference([NotNull] this ISolution solution)
         public static bool IsUnityProject([CanBeNull] this IProject project)
         {
             // Only VSTU adds the Unity project flavour. Unity + Rider don't, so we have to look at references
-            return project != null && (project.HasFlavour<UnityProjectFlavor>() || ReferencesUnity(project));
+            return project != null && (project.HasFlavour<UnityProjectFlavor>() || project.IsValid() && ReferencesUnity(project));
```
citizenmatt: Why would `project` be invalid here?

van800: missing/incompatible msbuild?

citizenmatt: Oh, would we still have a project in that case?

van800: I don't know, I can't reproduce it for real

PR #1109, opened 2019-03-27
Bug fixes

### LINKED ISSUE #1038: "String based property lookup is inefficient" suggestion is not displayed for MaterialPropertyBlock methods
Despite almost having the same methods as the class `Material`, when accessing properties of a `MaterialPropertyBlock` with strings, the warning is not displayed.

image

### LINKED ISSUE #1044: Highlight AddComponent(Type t) as expensive operation
Inside a performance critical context, Rider will currently highlight `gameObject.AddComponent<Animator>()` as an expensive operation, but it doesn't highlight `gameObject.AddComponent(typeof(Animator))`. 

These two calls are equivalent - the generic version is simply implemented as `AddComponent(typeof(T)) as T`.

### LINKED ISSUE #1083: NoNullPropagation spelled wrong as NoNullPropogation
There seems to be a spelling error for the warning `Unity.NoNullPropagation`. The warning is actually named `NoNullPropogation`. When you want to disable it once with a comment in Rider, it will then display a warning because of the wrong spelling.

https://github.com/JetBrains/resharper-unity/blob/1edad4e83421d388524f612f1929969af8579198/resharper/resharper-unity/src/CSharp/Daemon/Errors/CSharpErrors.xml#L504

krasnotsvetov:
1. Fix #1044 
2. Fix #1083 
3. Fix #1038 
4. Fix RIDER-26074
5. Fix RIDER-26108
6. Fix RIDER-26101
7. Unity dll project has same Unity pop-up menu
8. Unity Editor will not be started via debug button for Unity dll project
9. "Unity is off" is not showing after Unity Editor is launched via Unity Code Vision context menu
10. "Unity is off" -> "Unity is not running"


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/ui/UnityImportantActions.kt
```diff
@@ -13,6 +15,20 @@ class UnityImportantActions : DefaultActionGroup(), DumbAware {
             return
         }
 
+        e.presentation.isVisible = true
+        e.presentation.icon = UnityIcons.Actions.UnityActionsGroup
+    }
+}
+
+class UnityDllImportantActions : DefaultActionGroup(), DumbAware {
```
citizenmatt: Why add this as a new action? Why not just change the `update` of `UnityImportantActions` to handle dll projects as well?

krasnotsvetov: It has another set of actions inside popup. Also I don't know good API to hide actions and separator which are declared in xml

citizenmatt: Do the other actions still work in a dll project? If the solution lives inside a Unity project structure, we can still get a connection, so we should be able to show log items, start Unity and install the plugin.

krasnotsvetov: Other actions don't work in Dll project. But when dll project lives inside Unity project structure, we will show the basic menu and hide reduced dll menu

citizenmatt: OK, cool. Thanks!

PR #1111, opened 2019-03-28
Exception fixes

van800:


unity/EditorPlugin/PluginSettings.cs
```diff
@@ -191,6 +191,11 @@ private static void RiderPreferencesItem()
         if (EditorGUILayout.Toggle(new GUIContent("Rider is default editor"), PluginEntryPoint.Enabled))
         {
           EditorPrefsWrapper.ExternalScriptEditor = RiderPathInternal;
+
+          // make sure the plugin was initialized first.
+          // this can happen in case "Rider" was set as the default scripting app only after this plugin was imported.
+          PluginEntryPoint.Init();
```
krasnotsvetov: Are we really want to call Init method each time, when user opens settings?

krasnotsvetov: I missed that plugin have been already initialized , so that's ok :-)

PR #1112, opened 2019-03-29
fix RIDER-26176

van800:


van800:
We decided not to hack this without repro.

PR #1118, opened 2019-04-01
Include Unity asset files in usage count Code Vision metrics

citizenmatt:
This PR adds Unity asset files to SWEA, meaning usages are counted and displayed as part of Code Vision metrics.


krasnotsvetov:
What are you thinking about new code vision for yaml usages? It will help to discover yaml features
https://youtrack.jetbrains.com/issue/RIDER-26267

cc @kskrygan

citizenmatt:
AIUI, it would require a separate SWEA stage to count those usages. Not entirely sure it's worth it, if the main count of actual usages is now correct.

citizenmatt:
OK, I have to park this for now. SWEA's `CollectUsagesStageProcess` stage will cause all chameleons to be opened, which is obviously very bad for performance. There isn't a way to work around this - if it's included in SWEA, `CollectUsagesStageProcess` will open all chameleons. If it's not included in SWEA, it doesn't get any invisible file processing, so we can't even make our own stage and separate Code Vision widget.

See [RSRP-474142](https://youtrack.jetbrains.com/issue/RSRP-474142)

Issue #326, opened 2018-01-10
Warn user on editing unity-generated csproj in rider


van800:
Editing csproj/sln doesn't make any sense for unity generated projects.

van800:
* doesn't make sense

van800:
https://github.com/JetBrains/resharper-unity/issues/517

PR #1135, opened 2019-04-12
Add editor notification when editing Unity files

citizenmatt:
- [x] Show notification panel when editing known asset files. Fixes [RIDER-23145](https://youtrack.jetbrains.com/issue/RIDER-23145). Note that we can't block editing of these files, as that would also block refactorings. So this adds a warning banner for known Unity files.
- [x] Show notification panel when editing `.sln` and `.csproj` files. This partially addresses #326, but doesn't help if the user edits the project file via properties, or e.g. excluding a file through context menus.


van800:
I think IsUnityGeneratedProject should be used instead of IsUnityProject.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/ui/NonUserEditableEditorNotification.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    companion object {
+        val KEY = Key.create<EditorNotificationPanel>("non-user.editable.source.file.editing.notification.panel")
+    }
+
+    override fun getKey(): Key<EditorNotificationPanel> = KEY
+
+    override fun createNotificationPanel(file: VirtualFile, fileEditor: FileEditor, project: Project): EditorNotificationPanel? {
+
+        if (project.isUnityProject()) {
```
van800: IsUnityGeneratedProject should be used, I think

citizenmatt: Good point. I think maybe `isUnityProject` for any kind of asset (because they're always dangerous to edit by hand), and `isUnityGeneratedProject` for solution and project files?

van800: Yes, that would work.

PR #1137, opened 2019-04-12
fix several "Not Bound" exceptions. DEXP-453883, RIDER-26176

van800:


van800:
Stopped by review. Decided to remove grouping event and add more logging instead.

PR #1142, opened 2019-04-16
Show usage count on properties

citizenmatt:
Usage count was missing on properties used in YAML - the usage count was applied to the setter accessor, not the property itself.

Also adds a check to see if the processing should be interrupted.


resharper/resharper-unity/src/Yaml/Psi/Resolve/EventHandlerArgumentMode.cs
```diff
@@ -1,14 +1,23 @@
...
+    // Static methods are listed, but described in the Inspector as "missing" and are not invoked.
+    // Private methods are not listed, but are invoked.
+    // Methods with return types are not listed, and are not invoked (there is a type mismatch exception at runtime).
+    // Methods with too many parameters are marked as "missing" and not invoked.
     public enum EventHandlerArgumentMode
     {
-        Unknown = 0,
-        None = 1,
-        UnityType = 2,
+        EventDefined = 0,
```
krasnotsvetov: Could you add tests for this case? 
See src\test\kotlin\FindUsagesAssetTest.kt

PR #1167, opened 2019-05-04
Added UsageInspectionSuppresions for UnityEditor.Build.*

### LINKED ISSUE #686: Classes deriving from IPreprocessBuild and IPostprocessBuild should be marked implicitly used
Sadly, `[MeansImplicitUse]` doesn't work here

JurjenBiewenga:
Fixes #686 

All the interfaces in UnityEditor.Build implement the [IOrderedCallback](https://docs.unity3d.com/2017.4/Documentation/ScriptReference/Build.IOrderedCallback-callbackOrder.html) interface however I could not figure out how to make the callbackOrder property implicitly used.


citizenmatt:
Thanks for the contribution, unfortunately, this is not the correct approach. The `api.xml` file is intended to hold event functions, parsed from the "Messages" section of the Unity documentation. The methods listed in the file are marked as implicitly used by the `UsageInspectionsSuppressor`, but that is because they are event functions, not because this is how to mark them as in use. It also doesn't work with properties, only methods.

A more appropriate approach would be to add this directly to `UsageInspectionsSuppressor`, by adding explicit checks for methods and properties, by looking at the short name and verifying that they implement the appropriate interface. E.g. 

```cs
case IProperty when IsOrderedCallbackMember():
  flags = ImplicitUseKindFlags.Access;
  return true;
```

and something similar for methods.

The implementation check would probably need to check the method or property's `ShortName` to see if it's an interesting member, and then call `member.GetRootSuperMembers(false)` and for each returned member, call `rootMember.GetContainingType()` and check to see if that type is `UnityEditor.Build.IOrderedCallback`.

resharper/resharper-unity/src/CSharp/Daemon/UsageChecking/UsageInspectionsSuppressor.cs
```diff
@@ -15,6 +19,21 @@ public class UsageInspectionsSuppressor : IUsageInspectionsSuppressor
     {
         private readonly ILogger myLogger;
 
+        private readonly List<IClrTypeName> implicitlyUsedInterfaces = new List<IClrTypeName>()
```
JurjenBiewenga: I was debating whether to add this to KnownTypes however I couldn't think of a case where you would need the same types. Any opinions?

JurjenBiewenga:
This solution might not work with different Unity versions in this form. I'm not sure how to test that however, any suggestions?

citizenmatt:
Why do you think it might not work with other Unity versions?

resharper/resharper-unity/src/CSharp/Daemon/UsageChecking/UsageInspectionsSuppressor.cs
```diff
@@ -92,13 +121,83 @@ public bool SuppressUsageInspectionsOnElement(IDeclaredElement element, out Impl
                     // But it still should be checked if the field is ever accessed from the code.
                     flags = ImplicitUseKindFlags.Assign;
```
citizenmatt: Feel free to kill this comment. It's not really helpful.

resharper/resharper-unity/src/CSharp/Daemon/UsageChecking/UsageInspectionsSuppressor.cs
```diff
@@ -92,13 +121,83 @@ public bool SuppressUsageInspectionsOnElement(IDeclaredElement element, out Impl
...
                     return true;
-                
+
+                case IProperty property when IsImplicitlyUsedInterfaceProperty(property):
+                    flags = ImplicitUseKindFlags.Assign;
+                    return true;
+
```
citizenmatt: Merge this `case` with the one below: `case IProperty property when Is..() || Is()…:`

resharper/resharper-unity/test/data/CSharp/Daemon/UsageChecking/PreprocessBuildInterface01.cs
```diff
@@ -0,0 +1,19 @@
...
+    }
+}
+
+public class B : A, IPostprocessBuild
+{
+    public void OnPostprocessBuild(BuildTarget target, string path)
+    {
+        throw new System.NotImplementedException();
+    }
+}
```
citizenmatt: Add another case where the interface is being implicitly implemented via a base class, e.g.

```
public abstract class Base : IPostprocessBuild
{
  // …
  public abstract void OnPostProcessBuild();
}
```

and

```
public class Derived : Base
{
  public override void OnPostProcessBuild(…) { }
}
```

This should also help check if we can short circuit the type check.

resharper/resharper-unity/src/CSharp/Daemon/UsageChecking/UsageInspectionsSuppressor.cs
```diff
@@ -92,13 +121,83 @@ public bool SuppressUsageInspectionsOnElement(IDeclaredElement element, out Impl
...
+            foreach (var implicitlyUsedTypeName in myImplicitlyUsedInterfaces)
+            {
+                var type = TypeFactory.CreateTypeByCLRName(implicitlyUsedTypeName, method.Module).GetTypeElement();
+
+                if (type == null)
+                    return false;
+                
+                foreach (var overridableMemberInstance in method.GetRootSuperMembers())
+                {
+                    if (type.DerivesFrom(overridableMemberInstance.DeclaringType.GetClrName()))
```
citizenmatt: Does `DeclaringType` always return back the interface? If so, we can simplify this to just comparing type names, and skipping the method name check too.

We'll need to check that `DeclaringType` is always the interface, even in the case of a derived class that is overriding a method declared in a base class that implements one of our interesting interfaces.

resharper/resharper-unity/src/CSharp/Daemon/UsageChecking/UsageInspectionsSuppressor.cs
```diff
@@ -92,13 +121,83 @@ public bool SuppressUsageInspectionsOnElement(IDeclaredElement element, out Impl
...
+            foreach (var implicitlyUsedTypeName in myImplicitlyUsedInterfaces)
+            {
+                var type = TypeFactory.CreateTypeByCLRName(implicitlyUsedTypeName, property.Module).GetTypeElement();
+
+                if (type == null)
+                    return false;
+                
+                foreach (var overridableMemberInstance in property.GetRootSuperMembers())
+                {
+                    if (type.DerivesFrom(overridableMemberInstance.DeclaringType.GetClrName()))
```
citizenmatt: As above for methods. Can we short circuit this check?

citizenmatt:
Oh yeah, and can you add an "added" line in the `CHANGELOG.md`, `plugin.xml` and `resharper-unity.nuspec`, too, please?

JurjenBiewenga:
Where are the plugin.xml and the resharper-unity.nuspec located?

resharper/resharper-unity/src/CSharp/Daemon/UsageChecking/UsageInspectionsSuppressor.cs
JurjenBiewenga: Why wouldn't we need to check the method name?

citizenmatt:
[resharper/resharper-unity/src/resharper-unity.resharper.nuspec](https://github.com/JetBrains/resharper-unity/blob/192/resharper/resharper-unity/src/resharper-unity.resharper.nuspec) and [rider/src/main/resources/META-INF/plugin.xml](https://github.com/JetBrains/resharper-unity/blob/192/rider/src/main/resources/META-INF/plugin.xml)

resharper/resharper-unity/src/CSharp/Daemon/UsageChecking/UsageInspectionsSuppressor.cs
citizenmatt: We know that any method/property on these interfaces should be marked as implicitly used, so if the `overridableMemberInstance.DeclaringType` is the interface itself (and not a derived type), then the method has to belong to that interface, and will be implicitly used.

citizenmatt:
LGTM

PR #1176, opened 2019-05-10
Avoid coping EditorPlugin for Unity 2019.2+

van800:
Merge only after com.unity.ide.rider with corresponding changes lands

### DIFF
--- a/rider/src/main/resources/META-INF/plugin.xml
+++ b/rider/src/main/resources/META-INF/plugin.xml
@@ -139,9 +139,6 @@
         <action id="ShowUnityLogInRiderAction"
                 class="com.jetbrains.rider.plugins.unity.actions.ShowUnityLogInRiderAction"
                 text="Show Unity Log Window" />
-        <action id="InstallEditorPluginAction"
-                class="com.jetbrains.rider.plugins.unity.actions.InstallEditorPluginAction"
-                text="Install Unity Editor Plugin" />
       </group>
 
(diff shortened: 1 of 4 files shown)


SugoiDev:
Is `com.unity.ide.rider` 1.0.7 unreleased? I only see up to 1.0.6 (even in [bintray](https://bintray.com/unity/unity-staging/com.unity.ide.rider))

van800:
Not yet. Should appear in a couple of days.

van800:
1.0.7 is available on staging!

rider/src/main/resources/META-INF/plugin.xml
```diff
@@ -184,6 +181,10 @@
             use-shortcut-of="ShowIntentionActions">
       <add-to-group group-id="SolutionExplorerPopupMenu" anchor="last" />
     </action>
+
+    <action id="InstallEditorPluginAction"
+            class="com.jetbrains.rider.plugins.unity.actions.InstallEditorPluginAction"
+            text="Install Unity Editor Plugin" />
```
citizenmatt: Would it be better to leave this plugin where it was, but add some logic in `update` to hide it if Unity is 2019.2+? So it would still be available for older versions of Unity, but hidden when it's not needed.

van800: Someone from the test team told me that this action is confusing and is not actually useful for anyone. I started to think about that and decided to remove it from the Menu.

PR #1178, opened 2019-05-11
Fix errors in event handlers resolve

krasnotsvetov:
Fix RIDER-27445
Fix RIDER-27684

### DIFF
--- a/resharper/resharper-unity/src/UnityApi.cs
+++ b/resharper/resharper-unity/src/UnityApi.cs
@@ -129,22 +129,22 @@ namespace JetBrains.ReSharper.Plugins.Unity
         // scene. We don't have enough information to do this by name, so we'll mark all potential event handlers as
         // implicitly used by Unity
         // See https://github.com/Unity-Technologies/UnityCsReference/blob/02f8e8ca594f156dd6b2088ad89451143ca1b87e/Editor/Mono/Inspector/UnityEventDrawer.cs#L397
-        public bool IsPotentialEventHandler([CanBeNull] IMethod method)
+        public bool IsPotentialEventHandler([CanBeNull] IMethod method, bool isFindUsages = false)
         {
             if (method == null || !method.ReturnType.IsVoid())
                 return false;
 
             // Type.GetMethods() returns public instance methods only
-            if (method.GetAccessRights() != AccessRights.PUBLIC || method.IsStatic)
+            if (method.GetAccessRights() != AccessRights.PUBLIC && !isFindUsages|| method.IsStatic)
                 return false;
 
             return IsUnityType(method.GetContainingType()) &&
                    !method.HasAttributeInstance(PredefinedType.OBSOLETE_ATTRIBUTE_CLASS, true);
         }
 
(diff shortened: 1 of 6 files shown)


resharper/resharper-unity/src/UnityApi.cs
```diff
@@ -129,22 +129,22 @@ public bool IsInjectedField([CanBeNull] IField field)
...
         // See https://github.com/Unity-Technologies/UnityCsReference/blob/02f8e8ca594f156dd6b2088ad89451143ca1b87e/Editor/Mono/Inspector/UnityEventDrawer.cs#L397
-        public bool IsPotentialEventHandler([CanBeNull] IMethod method)
+        public bool IsPotentialEventHandler([CanBeNull] IMethod method, bool isFindUsages = false)
         {
             if (method == null || !method.ReturnType.IsVoid())
                 return false;
 
             // Type.GetMethods() returns public instance methods only
-            if (method.GetAccessRights() != AccessRights.PUBLIC || method.IsStatic)
+            if (method.GetAccessRights() != AccessRights.PUBLIC && !isFindUsages|| method.IsStatic)
```
citizenmatt: I think it's worth adding a comment here to say that the Unity Editor will only list public methods, but will invoke any method, even if it's private.

It might even be worth getting rid of the `isFindUsages`, and making `true` the default action.

resharper/resharper-unity/src/Yaml/Psi/Resolve/EventHandlerSymbolFilter.cs
```diff
@@ -0,0 +1,58 @@
...
+using JetBrains.ReSharper.Psi.Resolve;
+
+namespace JetBrains.ReSharper.Plugins.Unity.Yaml.Psi.Resolve
+{
+    public class EventHandlerSymbolFilter : SimpleSymbolFilter
+    {
+        private IType myType;
+        public EventHandlerSymbolFilter(int mode, string type, IPsiModule psiModule)
+        {
+            if (mode == 2)
```
citizenmatt: Can we use an enum or constants instead of raw values? I don't know what 0 or 1 means here.

resharper/resharper-unity/src/Yaml/Psi/Resolve/UnityEventTargetReferenceFactory.cs
```diff
@@ -62,7 +63,18 @@ public ReferenceCollection GetReferences(ITreeNode element, ReferenceCollection
                 var fileID = callMapNode.FindMapEntryBySimpleKey("m_Target")?.Value.AsFileID();
                 if (fileID != null && !fileID.IsNullReference && fileID.guid == null)
                 {
-                    var reference = new UnityEventTargetReference(methodNameValue, fileID);
+                    var text = callMapNode.Entries.FirstOrDefault(t => t.Key.MatchesPlainScalarText("m_Mode"))?.Value
+                        .GetPlainScalarText();
+                    if (!int.TryParse(text, out var mode))
+                        return ReferenceCollection.Empty;
```
citizenmatt: What happens if `m_Mode` is empty?

PR #1223, opened 2019-06-11
workaround for DEXP-468728

ArsenyChernyaev:


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/configurations/UnityAttachToEditorRunConfiguration.kt
```diff
@@ -103,6 +110,32 @@ class UnityAttachToEditorRunConfiguration(project: Project, factory: UnityAttach
         return true
     }
 
+    //get processes on background thread with progress
+    private fun getProcesses(): Array<ProcessInfo> {
+        val promise = AsyncPromise<Array<ProcessInfo>>()
```
kskrygan: if this is only on a background thread, please assert it.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/configurations/UnityAttachToEditorRunConfiguration.kt
```diff
@@ -103,6 +110,32 @@ class UnityAttachToEditorRunConfiguration(project: Project, factory: UnityAttach
...
+                    val processes = OSProcessUtil.getProcessList()
+                    promise.setResult(processes)
+                } catch (t: Throwable) {
+                    promise.setError(t)
+                }
+            }
+
+        }.queue()
+
+        if (!pumpMessages(WAIT_FOR_PROCESSES_TIMEOUT) {
```
kskrygan: I'm strictly against unnecessary pumping, please consider using spinUntil API

PR #1228, opened 2019-06-14
Search Unity installations installed by Unity in-house app

van800:


van800:
I am closing this in favor of different approach, which should always add Unity, found by references in the csproj to list of found Unity apps.

PR #1252, opened 2019-07-19
Remove Boo and UnityScript from import completion

### LINKED ISSUE #574: Sort System.Collections.Generic above Boo.Lang
image

Generally I'd imagine Generic.List<> should be prioritised, and in this case it's extremely frustrating to always get references to the former.

citizenmatt:
Adds `Boo.Lang.*` and `UnityScript.*` as namespaces to be excluded from auto-import.

Note that this means excluded from import completion lists, the auto-import popup and any other quick fixes. The only way to use types from these namespaces is to manually add the `using` statement.

Note that the current implementation might have to change before 192 is completed. The auto import settings added to the in memory solution settings layer currently completely override any in the global settings. This will hopefully be fixed before RTM - see [RIDER-30397](https://youtrack.jetbrains.com/issue/RIDER-30397)

Fixes #574


citizenmatt:
Paused until we get a better API for merging settings from plugins.

citizenmatt:
Maybe also consider adding `System.Diagnostics.Debug` to avoid clashes with `UnityEngine.Debug`. Must be careful that this only applies to Unity projects.

citizenmatt:
Closing in favour of #1473. Merge conflicts means there's nothing useful to keep from this PR.

PR #1259, opened 2019-07-26
Unchanged and serializable code vision, two new CA

krasnotsvetov:
1. Show unchanged and serializable code vision for fields without changes in Unity editor
2. CA for adding RequireComponent attribute
3. CA for initialization component field via `GetComponent` in `Start` or `Awake` method


resharper/resharper-unity/src/CSharp/Daemon/Stages/Highlightings/IconsProviders/FieldDetector.cs
```diff
@@ -39,24 +39,24 @@ public override IDeclaredElement Analyze(IDeclaration element, IHighlightingCons
             bool isSerializedField = myUnityApi.IsSerialisedField(declaredElement);
             if (isSerializedField)
             {
+                const string displayText = "Serializable";
                 if (myUnityApi.IsDescendantOfMonoBehaviour(declaredElement.GetContainingType()))
                 {
-                    AddMonoBehaviourHighlighting(consumer, field, "Serialized field", "This field is initialised from Inspector", kind);
+                    AddMonoBehaviourHighlighting(consumer, field, displayText, "This field is initialised from Inspector", kind);
```
van800: "This field is initialised from Inspector" to const

PR #1263, opened 2019-07-29
Show autosave configure notifcation after first typing in play mode

krasnotsvetov:


unity/EditorPlugin/AfterUnity56/Navigation/Initialization.cs
```diff
@@ -57,6 +59,32 @@ public static void OnModelInitializationHandler(UnityModelAndLifetime modelAndLi
...
       });
+      
+      modelValue.ShowPreferences.Advise(connectionLifetime, result =>
+      {
+        if (result != null)
+        {
+          MainThreadDispatcher.Instance.Queue(() =>
+          {
+
+            var tab = UnityUtils.UnityVersion >= new Version(2018, 2) ? "_General" : "Rider";
```
van800: adding try-catch-log may be good idea here.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/UnityAutoSaveConfigureNotification.kt
```diff
@@ -0,0 +1,128 @@
...
+                        if (!unityHost.model.scriptCompilationDuringPlay.hasValue)
+                            return
+
+                        if (unityHost.model.scriptCompilationDuringPlay.valueOrThrow != ScriptCompilationDuringPlay.RecompileAndContinuePlaying)
+                            return
+
+                        if (!lifetimeDefinition.isAlive)
+                            return
+
+                        val generalSettings = GeneralSettings.getInstance()
```
kskrygan: let's move GeneralSettings.getInstance out of this scope, so it won't be calculated for every doc change

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/UnityAutoSaveConfigureNotification.kt
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    init {
+        solutionLifecycleHost.isBackendLoaded.whenTrue(componentLifetime) {
+            if (!propertiesComponent.getBoolean(settingName) && unityProjectDiscoverer.isUnityProject) {
+
+                val eventMulticaster = EditorFactory.getInstance().eventMulticaster
+                val documentListener: DocumentListener = object : DocumentListener {
+                    override fun documentChanged(event: DocumentEvent) {
+                        if (unityHost.model.editorState.valueOrDefault(EditorState.Disconnected) != EditorState.ConnectedPlay)
```
kskrygan: if this works for o(1) then it's fine.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/UnityAutoSaveConfigureNotification.kt
```diff
@@ -0,0 +1,128 @@
...
+                            return
+
+                        if (!unityHost.model.scriptCompilationDuringPlay.hasValue)
+                            return
+
+                        if (unityHost.model.scriptCompilationDuringPlay.valueOrThrow != ScriptCompilationDuringPlay.RecompileAndContinuePlaying)
```
kskrygan: why valueOrThrow? Isn't throwing assertions here too aggressive?

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/UnityAutoSaveConfigureNotification.kt
```diff
@@ -0,0 +1,128 @@
...
+
+            val panel = EditorNotificationPanel(LightColors.RED)
+            panel.setText("You are modifying a script while Unity Editor is being in Play Mode. This can lead to a loss of the state in your running game.")
+
+            panel.createActionLabel("Configure Unity Editor") {
+                unityHost.model.showPreferences.fire()
+                lifetimeDefinition.terminate()
+            }
+
+            panel.createActionLabel("Don't show again") {
```
kskrygan: "Do not show again"
Please avoid using "don't", "ain't" and other reductions in UI

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/UnityAutoSaveConfigureNotification.kt
krasnotsvetov: Ooops. Thanks!

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/UnityAutoSaveConfigureNotification.kt
krasnotsvetov: AFAIK, our model properties cache value

PR #1268, opened 2019-08-05
Disable errors from shaders in swa in rtm

krasnotsvetov:


resharper/resharper-unity/src/ShaderLab/Daemon/Stages/IdentifierHighlightingStage.cs
```diff
@@ -15,17 +16,20 @@ public class IdentifierHighlightingStage : ShaderLabStageBase
     {
         private readonly ResolveHighlighterRegistrar myRegistrar;
         private readonly ConfigurableIdentifierHighlightingStageService myIdentifierHighlightingStageService;
+        private readonly RunsProducts.ProductConfigurations myProductConfigurations;
```
IvanPashchenko: w

PR #1291, opened 2019-08-15
optimize performance

van800:
image
image

### DIFF
--- a/unity/EditorPlugin/PluginEntryPoint.cs
+++ b/unity/EditorPlugin/PluginEntryPoint.cs
@@ -544,7 +544,7 @@ namespace JetBrains.Rider.Unity.Editor
       editorPluginModel.PlayerLogPath.SetValue(playerLogPath);
     }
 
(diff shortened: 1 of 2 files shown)


unity/EditorPlugin/PluginEntryPoint.cs
```diff
@@ -544,7 +544,7 @@ private static void InitEditorLogPath(EditorPluginModel editorPluginModel)
       editorPluginModel.PlayerLogPath.SetValue(playerLogPath);
     }
 
-    internal static readonly string LogPath = Path.Combine(Path.Combine(Path.GetTempPath(), "Unity3dRider"), $"{DateTime.Now:yyyy-MM-ddT-HH-mm-ss}.log");
```
citizenmatt: `DateTime.UtcNow` might help if you still want to use date. `Now` is slow because it has to look up timezone information.

van800: I tried that. It was still slow

PR #1299, opened 2019-08-23
Fix resolve error in yaml files

krasnotsvetov:


resharper/resharper-unity/src/Yaml/Psi/Resolve/MonoScriptReference.cs
```diff
@@ -69,7 +69,8 @@ public override ISymbolTable GetReferenceSymbolTable(bool useReferenceName)
             {
                 if (useReferenceName && resolvedName != null)
                 {
-                    if (resolvedName == candidate.ShortName)
+                    if (resolvedName == candidate.ShortName && !candidate.HasTypeParameters() 
+                        && candidate.GetContainingType() == null)
```
citizenmatt: The check for nested types is already handled in `UnityObjectPsiUtil.GetTypeElementCandidatesFromScriptAssetGuid`. It makes sense to put the type parameter check there, too, instead of here.

resharper/resharper-unity/src/CSharp/Daemon/Stages/Analysis/MultiplicationOrderAnalyzer.cs
```diff
@@ -63,6 +63,11 @@ public MultiplicationOrderAnalyzer(UnityApi unityApi)
                         break;
                     }
                 }
+                
+                // incomplete expression
+                if (scalar == null)
+                    return;
+
```
citizenmatt: This doesn't match the PR title. If this is fixing a YT issue, can it be added to `CHANGELOG.md`, `plugin.xml` and `resharper-unity.nuspec`?

PR #1309, opened 2019-09-03
Fix exception when opening project with hidden Unity Explorer

citizenmatt:
Fixes [DEXP-482252](https://youtrack.jetbrains.com/issue/DEXP-482252)


citizenmatt:
Fix to changes from #1303, didn't go public, no need to added to CHANGELOG

PR #1319, opened 2019-09-14
Add Pure annotations for profiler struct

citizenmatt:
Prevents "Possibly impure struct method called on readonly variable" warnings


citizenmatt:
Closing as obsolete. Looks like the APIs are annotated in Unity's assemblies

PR #1322, opened 2019-09-20
Fix RIDER-32498 Rider synchronize project all the time

van800:


van800:
Fixed on Rider side (2019.3)

PR #1344, opened 2019-10-04
Focus/expand Unity windows, when Rider is showing usages in it

van800:


citizenmatt:
Could you add something to `CHANGELOG.md` and `plugin.xml` with change details before merging please?

PR #1362, opened 2019-10-14
Settings provider usage inspection

### LINKED ISSUE #1225: SettingsProvider is marked unused
In the code below both PluginSettings and RiderPreferencesItem are marked as unused.

```
  public class PluginSettings
  {
    [SettingsProvider]
    private static SettingsProvider RiderPreferencesItem()
    {
}}
```

https://github.cds.internal.unity3d.com/unity/com.unity.ide.rider/blob/master/Packages/com.unity.ide.rider/Rider/Editor/PluginSettings.cs#L141

JurjenBiewenga:
Currently there are 2 issues.

The test can't find `UnityEditor.SettingsEditor` and checking whether the return type is the correct type can most likely be done in a better way.

This closes #1225


resharper/resharper-unity/src/CSharp/Daemon/UsageChecking/UsageInspectionsSuppressor.cs
```diff
@@ -128,6 +135,20 @@ public bool SuppressUsageInspectionsOnElement(IDeclaredElement element, out Impl
             return false;
         }
 
+        private static bool IsSettingsProvider(IMethod method)
+        {
+            if (method.HasAttributeInstance(KnownTypes.SettingsProviderAttribute, AttributesSource.All) && method.IsStatic)
+            {
+                if (method.ReturnType.IsImplicitlyConvertibleTo(TypeFactory.CreateTypeByCLRName(KnownTypes.SettingsProvider, method.Module),
+                    new XamlWinRTTypeConversionRule(method.Module)))
```
krasnotsvetov: XamlWinRTTypeConversionRule is incorrect here, please use `method.Module.GetTypeConversionRule()`

citizenmatt: You should be able to do something like:

```
(method.ReturnType as IDeclaredType)?.GetTypeElement()?.DerivesFrom(clrTypeName);
```

Although possibly not all on one line :)

PR #1366, opened 2019-10-17
fixed early lifetime termination of process listener

ArsenyChernyaev:


ArsenyChernyaev:
See [this issue](https://youtrack.jetbrains.com/issue/RIDER-34039) for details

van800:
It may be similar to https://github.com/JetBrains/resharper-unity/pull/1298
@citizenmatt please take a look

citizenmatt:
Yes, this is fixed as part of #1298.

Issue #1231, opened 2019-06-17
Removing file with Safe Delete does not remove corresponding .meta file


B-Reif:
I removed a class with Safe Delete and allowed Rider to delete the resulting empty file. Afterward Unity warned me about deleting the corresponding .meta file.

Encountered this issue on Rider 2019.1.2.

Issue #1370, opened 2019-10-23
"Move to folder" refactoring doesn't move meta files


mnarimani:
Moving classes using Rider's "Move to folder" refactoring option doesn't move meta file of that class.
That causes "missing script" error on game objects.

van800:
Rider version? Unity version? Os?

mnarimani:
Rider 2019.2.3
Unity 2019.2.3f1
Rider Plugin 1.1.0
Windows 10

van800:
Unrelated to request, just want to let you know that
Rider Plugin 1.1.0 is not a good one.
Use at least 1.1.1 with `Test Framework` package 1.1.1.
Or 1.1.2-preview / 1.1.3-preview, both don't have requirement to have `Test Framework`.

I will try to reproduce the `Move to folder` request. Thank you.

van800:
This issue doesn't reproduce for me. There must be something special in your case. Does it reproduce in simple boiler-plate project?

van800:
Please provide more information.

mnarimani:
Sorry I totally forgot about this.
I still can reproduce it in an empty project. Here is a video of this bug:
https://youtu.be/Zx5tGSNfhHY

van800:
Reproduced. Thank you!

Issue #856, opened 2018-10-22
Renaming classes in Packages fails to maintain GUIDs


vertxxyz:
When refactoring the names of scripts through Rider metadata is usually maintained, but this stops seeming to be the case when scripts are in a Package directory.

PR #1389, opened 2019-11-01
Fix moving meta file for Move to Folder refactoring invoked on class

van800:
https://github.com/JetBrains/resharper-unity/issues/1370


resharper/resharper-unity/src/ProjectModel/MetaFileTracker.cs
```diff
@@ -210,7 +206,13 @@ private void OnItemRemoved(ProjectItemChange change)
                     RenameMetaFile(metaFile, newMetaFile, " via add/remove");
                 }
                 else
+                {
+                    // Only delete the meta file if the original file or folder is missing
+                    if (change.OldLocation.Exists != FileSystemPath.Existence.Missing)
+                        return;
+
```
citizenmatt: We can just get rid of this whole check. The old file will always still exist AFAICT, because we remove it from the project model, send out the notifications, and then remove from disk.

Leaving this check in means that a safe delete action will fail - we've removed the file from the project model, we get the notification, the file still exists on the disk and so we fail to delete the `.meta` file.

citizenmatt: So that would also fix #1231

citizenmatt: So that would also fix #1231

citizenmatt:
I think we could also get rid of the check for `IsAsset` - we already checking `project.IsUnityProject()` earlier in the process, so I think any change we get needs to be processed. Also, it means that any changes under `Packages` will fail to update the `.meta` files (#856) and I suspect any packages added with `file:` and pointing out of the normal project folder structure will have a good chance of failing too.

van800:
With all the proposed changes I start seeing https://youtrack.jetbrains.com/issue/DEXP-486193
@krasnotsvetov What do you think?

PR #1397, opened 2019-11-09
[203] Rename protocol models

citizenmatt:
The current naming of the protocol models is a little confusing - `RdUnityModel` doesn't talk to Unity and `EditorPluginModel` has nothing to do with Rider's editor. This PR renames the models too make it more obvious what they do, and who they talk to.

- [x] Rename `RdUnityModel` to `FrontendBackendModel`. This talks between the frontend and the backend
- [x] Rename `EditorPluginModel` to `BackendUnityModel`. This talks between the backend and the Unity editor plugin
- [x] Move the models to the `JetBrains.Rider.Model` and `com.jetbrains.rider.model` namespaces, as used by the other models in Rider
- [x] Remove the use of the "backend <-> Unity" model from the frontend. This was so that the frontend/backend protocol could reuse enums from the backend/Unity. This has been replaced with extra enums and explicit mapping
- [x] Provide backwards compatibility to workaround breakages due to renaming the backend/Unity protocol. Because models are keyed on type name, changing the name will mean that the protocol connects but the models don't sync, and nothing works. Creating an empty derived class in both the backend and the Unity editor allows using `BackendUnityModel` in code, but keeps the compatibility with the existing `EditorPluginModel`. This works for new Rider with old Unity plugin and new Unity plugin with old Rider.
- [ ] Tidy up some of the properties in the models (group by functionality, common naming style, etc.)


van800:
I removed 
- [ ] Warn if models are out of sync when loaded by the package. Currently, we only warn if the option to automatically update the plugin in Assets is disabled.
because I am going to address it in https://youtrack.jetbrains.com/issue/RIDER-35297

citizenmatt:
Paused. Will pick up again at the start of the 2020.2 cycle

citizenmatt:
Obsolete. Reimplemented in #1887

PR #1443, opened 2019-12-09
Make CSS an optional dependency

citizenmatt:
We use the CSS tooling to provide USS support. But if CSS is disabled in plugins, the entire Unity plugin is also disabled. This PR makes the CSS support an optional dependency. See [RIDER-36523](https://youtrack.jetbrains.com/issue/RIDER-36523).

* If CSS is enabled, everything works as before.
* If CSS is disabled, there is no syntax highlighting or other USS support, but the rest of the Unity plugin continues to work.
* The Unity plugin will also show an editor notification for `*.uss` files if CSS is disabled, prompting to re-enable the plugin, and show a toast notification to restart.

This PR also adds the correct icons for `*.uss` and `*.uxml` files (see [RIDER-34788](https://youtrack.jetbrains.com/issue/RIDER-34788))


citizenmatt:
Feel free to comment on wording, etc.

Screenshot 2019-12-09 at 14 57 02

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/util/FileExtensions.kt
```diff
@@ -52,4 +52,5 @@ fun isGeneratedUnityFile(file: VirtualFile): Boolean {
     return extension.equals("csproj", true) || extension.equals("sln", true)
 }
 
-fun isUxmlFile(file: VirtualFile) = file.extension.equals("uxml", true)
\ No newline at end of file
+fun isUxmlFile(file: VirtualFile) = file.extension.equals("uxml", true)
```
kashmervil: it's a better practice to operate with file types rather than with actual hard-coded strings comparison. custom file associations can be added, there could be another extension in the future etc. 

This is not a big problem for your scenario, and you also covered the 'ignore case' thing, but still...

Consider using `FileTypeRegistry#isFileOfType` is possible

kashmervil:
👍

PR #1473, opened 2019-12-22
Exclude conflicting type names from import completion

### LINKED ISSUE #574: Sort System.Collections.Generic above Boo.Lang

citizenmatt:
Adds `Boo.Lang.*` and `UnityScript.*` as namespaces to be excluded from auto-import. Also adds `System.Diagnostics.Debug` as a method to be ignored. This means that these namespaces and types are not shown in import completion popup lists, the auto-import tooltip popup, and any other quick fix. The only way to use these types is to manually add the using statement.

This addresses the issue that trying to complete `List<T>` can match `Boo.Lang.List<T>`.

This PR also adds an internal action to Rider that will dump all duplicate type names in a given project. Useful to show candidates that could use further rules.

Fixes #574

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ This plugin has functionality that is common to both ReSharper and Rider. It als
 * [Commits](https://github.com/JetBrains/resharper-unity/compare/net193...net201)
 * [Milestone](https://github.com/JetBrains/resharper-unity/milestone/32?closed=1)
 
+### Added
+
+- Exclude `Boo` and `UnityScript` namespaces, as well as the `System.Diagnostics.Debug` type from import completion ([#574](https://github.com/JetBrains/resharper-unity/issues/574), [#1473](https://github.com/JetBrains/resharper-unity/pull/1473))
+
 
 
 ## 2019.3.1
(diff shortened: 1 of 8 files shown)


SugoiDev:
Ignoring `System.Diagnostics.Debug` will help a lot, since it is often the first choice when you use a `Debug.Log*` method and is always the wrong one.

I would really like to see this generalized to allow the user to add a new import completion ignore list.
There are quite a few DLLs in third party packages that, despite having public methods, should never be used. In particular obfuscated ones that only add noise.

citizenmatt:
This is already a generalised mechanism. This PR will add default entries into the _Preferences | Editor | General | Auto Import_ settings page. You're free to add your own.

citizenmatt:
Other candidates for exclusion:

* `System.Numerics.Vector2`, `Vector3`, `Vector4`, `Quaternion` and `Plane` conflict with the same types in `UnityEngine`. I think this is probably worth adding. In the context of Unity, you'll want to use the Unity types, as that's what the Unity APIs will expect.
* `System.Random` and `UnityEngine.Random`. There is an argument for using `System.Random` - it's an object, so you can create a separate random source for different consumers, you can create an instance with a given feed so you can make it a little more reproducible, etc. But you can also argue those are advanced use cases, so it's ok to make you add the `using` manually.

Comments?

citizenmatt:
Merged as-is. We'll review `System.Numerics` and `Random` during the EAP.

citizenmatt:
See also [RIDER-33445](https://youtrack.jetbrains.com/issue/RIDER-33445)

keith-miller:
Thank you so much for this!

PR #1521, opened 2020-02-12
Fix building searchable options (net193)

citizenmatt:
Backport #1520 to `net193` in case there's another 2019.3 maintenance release


citizenmatt:
Oh. This PR was meant to be against `net193`, but was applied to `net201`. Fortunately, the changes had already been made, so this was essentially a nop. At least there's no need to revert.

PR #1551, opened 2020-03-03
Cleanup NoWarn 0169 from Unity generated project

van800:
related to https://github.com/JetBrains/resharper-unity/issues/1552

### DIFF
--- a/unity/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
+++ b/unity/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
@@ -149,8 +149,14 @@ namespace JetBrains.Rider.Unity.Editor.AssetPostprocessors
       changed |= SetXCodeDllReference("UnityEditor.iOS.Extensions.Common.dll", projectContentElement, xmlns);
       changed |= SetDisableHandlePackageFileConflicts(projectContentElement, xmlns); // already exists
       changed |= SetGenerateTargetFrameworkAttribute(projectContentElement, xmlns); // no need
+      changed |= CleanUpNoWarn(projectContentElement, xmlns); 
       
       return changed;
+    }
+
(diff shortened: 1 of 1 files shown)


unity/EditorPlugin/AssetPostprocessors/CsprojAssetPostprocessor.cs
```diff
@@ -149,10 +149,16 @@ private static bool UpgradeProjectFile(string projectFile, XDocument doc)
...
       changed |= SetDisableHandlePackageFileConflicts(projectContentElement, xmlns); // already exists
       changed |= SetGenerateTargetFrameworkAttribute(projectContentElement, xmlns); // no need
+      changed |= CleanUpNoWarn(projectContentElement, xmlns); 
       
       return changed;
     }
 
+    private static bool CleanUpNoWarn(XElement projectContentElement, XNamespace xmlns)
+    {
+      return SetOrUpdateProperty(projectContentElement, xmlns, "NoWarn", existing => "");
```
citizenmatt: We should only remove `0169`, in case Unity add another warning in the future. I think they're semi-colon separated, but would have to check - it might support semi-colon and comma as a separator.

van800: This would not change. After 2019.2 all ProjectGeneration things are on the Rider package side. So I don't expect any surprises.

citizenmatt: Good point! 👍

citizenmatt:
Is this obsolete now? Unity doesn't generate this `NoWarn` any more, does it?

van800:
> Is this obsolete now? Unity doesn't generate this `NoWarn` any more, does it?

It targets pre-2019.2.

van800:
No-one asked for this for a long time, so it would be wise to avoid touching long-time-working functionality.

PR #1564, opened 2020-03-11
Improve failed to attach message

van800:
fix https://youtrack.jetbrains.com/issue/RIDER-40881


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/UnityDebuggerOutputListener.kt
```diff
@@ -2,30 +2,55 @@ package com.jetbrains.rider.plugins.unity.run
...
+                    text += "\nPlease ensure that the player has 'Script Debugging' enabled and that the host '$host' is reachable.\n"
+                }
+
+            val debugNotification = XDebuggerManagerImpl.NOTIFICATION_GROUP.createNotification(text, NotificationType.ERROR)
+
+            debugNotification.setListener { notification, hyperlinkEvent ->
+                if (hyperlinkEvent.eventType != HyperlinkEvent.EventType.ACTIVATED)
+                    return@setListener
+
+                BrowserUtil.browse("https://docs.unity3d.com/$unityVersion/Documentation/Manual/ManagedCodeDebugging.html")
```
citizenmatt: Can this by `hyperlinkEvent.getURL()`? Would be nice to just use the URL you've already specified rather than have it in 3 different places.

PR #1565, opened 2020-03-11
WIP: Call-graph based Unity.Burst support

Domonion:
Support for Unity.Burst
1. highlights access to managed objects, object creation etc
2. transitively propagates burst contexts 
3. highlighting are taken and annotated to match Burst compiler


Domonion:
https://github.com/JetBrains/resharper-unity/pull/1665

Issue #1571, opened 2020-03-16
PlayMode tests always inconclusive


VacuumBreather:
While EditMode tests run fine inside Rider, the PlayMode tests always become inconclusive/not run.
While this happens, the Test Runner in Unity shows the editmode tests in the playmode tab, which is weird. If I "run all" in the Unity Test runner, that view resets to the correct tests again.

JetBrains Rider 2019.3.4
Build #RD-193.6494.48, built on February 27, 2020

[rider-logs-20200316-103147.zip](https://github.com/JetBrains/resharper-unity/files/4337319/rider-logs-20200316-103147.zip)

van800:
What is your Unity version?

VacuumBreather:
Version 2019.3.1f1

van800:
Updating Rider package  to the latest via Unity package manager should help.

VacuumBreather:
I already did that. The package version is 2.0.1

van800:
Have you switched this drop-down in Rider?
![image](https://user-images.githubusercontent.com/1482681/76750297-4a669b00-677e-11ea-9303-547fdd26c183.png)

VacuumBreather:
I have no idea where this button is supposed to be:
![screenshot](https://user-images.githubusercontent.com/61427715/76770289-f8357200-679d-11ea-9e82-1c1a6a98f5d9.jpg)

van800:
That's weird.
![image](https://user-images.githubusercontent.com/1482681/76770472-4f3b4700-679e-11ea-8749-1c73c9ac9b64.png)

van800:
Could you please show me the screenshot of the status-bar?
![image](https://user-images.githubusercontent.com/1482681/76770850-d092d980-679e-11ea-9c7f-491b84a18352.png)

VacuumBreather:
Here you go:
![screenshot2](https://user-images.githubusercontent.com/61427715/76771531-c7eed300-679f-11ea-9257-bf80950f203a.jpg)

van800:
Your project was not recognized as a Unity generated one.

VacuumBreather:
Ah... I opened a Visual Studio Solution filter instead of the solution file itself. Is that the issue? That's kind of a shame. I would rather not see all those third party projects all the time...

van800:
What is that `filter`? Could you please make a screenshot?

VacuumBreather:
[...]

PR #1577, opened 2020-03-17
Consider slnf be a valid solution file extension

van800:
https://github.com/JetBrains/resharper-unity/issues/1571

### DIFF
--- a/resharper/resharper-unity/src/ProjectModel/UnitySolutionTracker.cs
+++ b/resharper/resharper-unity/src/ProjectModel/UnitySolutionTracker.cs
@@ -39,9 +39,7 @@ namespace JetBrains.ReSharper.Plugins.Unity.ProjectModel
         private void SetValues()
         {
             IsUnityProjectFolder.SetValue(HasUnityFileStructure(mySolution.SolutionDirectory));
-            IsUnityProject.SetValue(IsUnityProjectFolder.Value && mySolution.IsValid() &&
-                                    (mySolution.SolutionFilePath.ExtensionNoDot.ToLower() == "sln"
-                                    ||mySolution.SolutionFilePath.ExtensionNoDot.ToLower() == "slnf"));
+            IsUnityProject.SetValue(IsUnityProjectFolder.Value && mySolution.IsValid() && mySolution.SolutionFilePath.ExistsFile);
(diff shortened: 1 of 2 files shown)


resharper/resharper-unity/src/ProjectModel/UnitySolutionTracker.cs
```diff
@@ -40,7 +40,8 @@ private void SetValues()
         {
             IsUnityProjectFolder.SetValue(HasUnityFileStructure(mySolution.SolutionDirectory));
             IsUnityProject.SetValue(IsUnityProjectFolder.Value && mySolution.IsValid() &&
-                                    mySolution.SolutionFilePath.ExtensionNoDot.ToLower() == "sln");
+                                    (mySolution.SolutionFilePath.ExtensionNoDot.ToLower() == "sln"
+                                    ||mySolution.SolutionFilePath.ExtensionNoDot.ToLower() == "slnf"));
```
citizenmatt: Is this check just to make sure we don't have a folder open? If so, can we simplify it to check that the solution isn't just a folder, and remove the explicit file extension check?

rider/src/main/kotlin/com/jetbrains/rider/UnityProjectDiscoverer.kt
```diff
@@ -55,7 +55,8 @@ class UnityProjectDiscoverer(project: Project) : LifetimedProjectService(project
     // Returns false when opening a Unity project as a plain folder
     private fun isCorrectlyLoadedSolution(project: Project): Boolean {
         val solutionFile = project.solutionFile
-        return project.solutionDescription is RdExistingSolution && solutionFile.isFile && solutionFile.extension.equals("sln", true)
+        return project.solutionDescription is RdExistingSolution && solutionFile.isFile &&
+            (solutionFile.extension.equals("sln", true) || solutionFile.extension.equals("slnf", true))
```
citizenmatt: Same as above. Is there a scenario when we get `RdExistingSolution` and `solutionFile` is a file, not a folder, and the file extension is NOT `sln` or `slnf`?

van800: Just checking `RdExistingSolution` is enough.

PR #1578, opened 2020-03-18
Marks providers fix

Domonion:
Marks provider must work at any tree node regardless if containing function null or not because some marks can be added outside of any function.


Domonion:
not relevant

PR #1592, opened 2020-03-26
Fix Refresh before running tests

van800:
+ fix Refresh before tests
+ may also fix RIDER-40852


unity/EditorPlugin/PluginEntryPoint.cs
```diff
@@ -414,7 +414,18 @@ private static void AdviseRefresh(EditorPluginModel model)
     {
       model.Refresh.Set((l, force) =>
       {
-        var task = new RdTask<Unit>();
+        var refreshTask = new RdTask<Unit>();
+        void SendResult()
+        {
+          if (!EditorApplication.isCompiling)
+          {
+            // ReSharper disable once DelegateSubtraction
```
krasnotsvetov: consider https://answers.unity.com/questions/14709/can-i-get-a-callback-when-scripts-have-finished-co.html

resharper/resharper-unity/src/Rider/UnitTesting/RunViaUnityEditorStrategy.cs
```diff
@@ -428,31 +441,38 @@ private void SubscribeResults(IUnitTestRun firstRun, Lifetime connectionLifetime
...
-            waitingLifetimeDef.Lifetime.OnTermination(() => tcs.TrySetCanceled());
-
+            var waitingLifetime = waitingLifetimeDef.Lifetime;
+            waitingLifetime.OnTermination(() => tcs.TrySetCanceled());
             tcs.Task.ContinueWith(_ => waitingLifetimeDef.Terminate(), lifetime);
+            if (cancellationToken.IsCancellationRequested) tcs.TrySetCanceled();
 
-            mySolution.Locks.ExecuteOrQueue(lifetime, "WaitForUnityEditorConnectedAndIdle", () =>
+            // Can't use ConnectionTracker because it pauses without focus
+            waitingLifetime.StartMainUnguardedAsync(async () =>
```
krasnotsvetov: @kskrygan, I am not sure about this code (while loop inside async main activity), could you take a look?

unity/EditorPlugin/PluginEntryPoint.cs
van800: doesn't fit.

resharper/resharper-unity/src/Rider/UnitTesting/RunViaUnityEditorStrategy.cs
kskrygan: @van800 @krasnotsvetov I would suggest using 
JetBrains.Application.Threading.Tasks.TaskHostEx.QueueRecuring
instead of a while loop. Please rework this part

van800: QueueRecuring only fires, when Rider has focus. While after running tests, some users what to do something else.

PR #1599, opened 2020-03-31
Safely remove all empty folders created by older Rider version

van800:
together with corresponding meta files


resharper/resharper-unity/src/Rider/UnityPluginInstaller.cs
```diff
@@ -114,12 +114,29 @@ private void InstallPluginIfRequired()
...
                     {
-                        myLogger.Info($"Remove {installationInfoToRemove.PluginDirectory}. Rider package should be used instead.");
-                        installationInfoToRemove.PluginDirectory.Delete();
+                        myLogger.Info($"Remove {pluginDll}. Rider package should be used instead.");
+                        pluginDll.DeleteFile();
+                        FileSystemPath.Parse(pluginDll.FullPath + ".meta").DeleteFile();
+                        
+                        // jetbrainsDir is usually "Assets\Plugins\Editor\JetBrains", however custom locations were also possible   
+                        var jetbrainsDir = installationInfoToRemove.PluginDirectory;
+                        if (jetbrainsDir.GetChildren().Any()) return;
```
citizenmatt: It's worth checking that the folders we're about to delete are the expected folders, so maybe check the folder name, too? Or just delete all empty folders up to `Assets`?

van800: Ok, I have added check for each folder name. We would not delete any folder with unexpected name.

PR #1607, opened 2020-04-03
[WIP] Debug UnityEditor startup

van800:
It works, however I need some dynamic port here.
Requires more effort. We need a new run-configuration for this.
I also think we need to kill Unity with Stop of this run-configuration.

### DIFF
--- a/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/configurations/UnityAttachToEditorProfileState.kt
+++ b/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/configurations/UnityAttachToEditorProfileState.kt
@@ -79,12 +79,14 @@ class UnityAttachToEditorProfileState(private val remoteConfiguration: UnityAtta
                         addPlayModeArguments(args)
                     }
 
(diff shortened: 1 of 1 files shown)


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/configurations/UnityAttachToEditorProfileState.kt
```diff
@@ -79,12 +78,13 @@ class UnityAttachToEditorProfileState(private val remoteConfiguration: UnityAtta
                         addPlayModeArguments(args)
                     }
 
-                    val process = ProcessBuilder(args).start()
-                    val actualPid = OSProcessUtil.getProcessID(process)
-                    remoteConfiguration.pid = actualPid
-                    remoteConfiguration.port = convertPidToDebuggerPort(actualPid)
+                    val p = 50013
```
ArsenyChernyaev, line 81: Probably we should call NetUtils.findFreePort here

van800:
code added as comment to https://github.com/JetBrains/resharper-unity/pull/1708

PR #1611, opened 2020-04-06
Update OnOpenAssetHandler.cs

PotterDai:
This solves the issue that when both Unity3D and Rider are in fullscreen mode (different workspaces) in MacOS, open script will jump back to desktop and then to Rider.


van800:
Thank you!

PR #1614, opened 2020-04-08
Stop trying to use MSBuild 15 without .NET Core SDK

ForNeVeR:
By default, our build system will try to use any instance of MSBuild 15 if it's found.

I have VS2017 installed for native development, but had no .NET development tools installed. So it tried to use older MSBuild that has no support of SDK-style projects, and failed to build.

We could fix that by adding explicit requirement for .NET Core SDK support.


ForNeVeR:
Well, something's definitely broken in my environment, but it was working on the CI before the patch. Let's put that aside for now, I guess. I'll apply the patch locally.

PR #1616, opened 2020-04-08
Fix resolving git packages for 2019.3+

citizenmatt:
Git based packages are not resolving properly in Unity 2019.3+. The format of the name of the cached package folder has changed - in Unity 2019.2, it's `{name}@{hash}`, while in 2019.3+, it only uses the first 10 characters of the hash.


zsavely:
@citizenmatt @krasnotsvetov 
It seems that on Unity version 2019.4.3 and Rider 2020.1.4 the bug has resurfaced. Can you confirm?

citizenmatt:
Yes, confirmed. Unity have moved the fit hash out of `manifest.json` and into a new file - `packages-lock.json`. I expect to get the fix into Rider 2020.2. See also [RIDER-47191](https://youtrack.jetbrains.com/issue/RIDER-47191)

PR #1629, opened 2020-04-21
Avoid calling Refresh on every save in the editor

van800:
Saves in the editor might be very frequent.
And each refresh may break completion https://youtrack.jetbrains.com/issue/RIDER-37420

From user logs:
image


van800:
reviewed by @krasnotsvetov

PR #1631, opened 2020-04-22
move FileSystemWatcher initialization to background thread

van800:


resharper/resharper-unity/src/Rider/UnityEditorProtocol.cs
```diff
@@ -91,21 +91,25 @@ public class UnityEditorProtocol
...
+                    // Add event handlers.
+                    watcher.Changed += OnChanged;
+                    watcher.Created += OnChanged;
 
-                lf.Bracket(() => { }, () =>
-                {
-                    watcher.Dispose();
+                    lf.Bracket(() => { }, () =>
+                    {
+                        watcher.Dispose();
```
epeshk, line 107: `lf.OnTermination(() => watcher.Dispose());` ?

PR #1640, opened 2020-04-30
Fix auto-detected LangVersion on Linux

van800:
also remove backslash, fix debug-only exception on linux

### DIFF
--- a/resharper/resharper-unity/src/UnityInstallationFinder.cs
+++ b/resharper/resharper-unity/src/UnityInstallationFinder.cs
@@ -284,11 +284,15 @@ namespace JetBrains.ReSharper.Plugins.Unity
                 Assertion.Assert(path.FullPath.EndsWith(".app", StringComparison.OrdinalIgnoreCase),
                     "path.FullPath.EndsWith('.app', StringComparison.OrdinalIgnoreCase)");
             }
-            else
+            else if (PlatformUtil.RuntimePlatform == PlatformUtil.Platform.Windows)
             {
                 Assertion.Assert(path.ExistsFile, "path.ExistsFile");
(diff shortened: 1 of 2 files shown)


resharper/resharper-unity/src/UnityInstallationFinder.cs
```diff
@@ -284,12 +292,16 @@ private static void AssertApplicationPath(FileSystemPath path)
...
+            else if (PlatformUtil.RuntimePlatform == PlatformUtil.Platform.Windows)
             {
                 Assertion.Assert(path.ExistsFile, "path.ExistsFile");
                 Assertion.Assert(path.FullPath.EndsWith(".exe", StringComparison.OrdinalIgnoreCase),
                     "path.FullPath.EndsWith('.exe', StringComparison.OrdinalIgnoreCase)");
             }
+            else
+            {
+                Assertion.Assert(path.ExistsFile, "path.ExistsFile");
+            }
```
citizenmatt, line 304: Let's make this a switch now. It should have been one to begin with.

resharper/resharper-unity/src/UnityInstallationFinder.cs
```diff
@@ -253,6 +253,14 @@ public static FileSystemPath GetAppPathByDll(XmlElement documentElement)
...
                         return exePath;
                 }
+                if (PlatformUtil.RuntimePlatform == PlatformUtil.Platform.Linux)
+                {
+                    var exePath = filePath.Combine("../../../Unity"); // Editor\Data\Managed\UnityEngine.dll
+                    if (!exePath.ExistsFile)
+                        exePath = filePath.Combine("../../../../Unity"); // Editor\Data\Managed\UnityEngine\UnityEngine.dll
+                    if (exePath.ExistsFile)
+                        return exePath;
+                }
```
citizenmatt, line 263: Convert to switch here too.

PR #1644, opened 2020-05-01
Correctly generate static constructor without body

citizenmatt:
Fixes an issue where the "redundant `[InitializeOnLoad]`" quick fix would fail to generate a static constructor if member generation settings were set to "default return value".


JurjenBiewenga:
lgtm

PR #1649, opened 2020-05-06
Convert quick fixes to bulk actions

citizenmatt:
The following Quick Fixes have been converted into bulk (scoped) actions, that work across method, class, file, folder, project and solution:

* Convert to `CompareTag`
* Convert to generic overload (`GameObject.GetComponent`, `GameObject.AddComponent`, `ScriptableObject.CreateInstance`)
* Combine setting parent with object creation ("Use Instantiate with parent argument")
* Convert coalescing operator to conditional
* Remove redundant attribute

The remove redundant attribute quick fix is now a "redundant code" quick fix and has additional features/implications:

* The quick fix is executed as part of removing redundant code in Code Cleanup
* Adds "remove redundant code in XXX" bulk fixes, which remove ALL redundant code, such as unnecessary `this.`, dead code, etc.
* The bulk quick fix will remove all redundant Unity attributes in scope, not just the attribute type under the caret, so it will remove any incorrectly applied `[SerializeField]`, `[InitializeOnLoad]`, `[HideInInspector]`, etc.


citizenmatt:
Another implication: if a non-serialisable base class has a `[SerializeField]` attribute, we currently consider that as redundant - we don't check for derived classes that might be serialised. The remove redundant attribute bulk fix will remove this attribute, even though it is valid.

This needs to become a show-stopper: [RIDER-18729](https://youtrack.jetbrains.com/issue/RIDER-18729)

resharper/resharper-unity/src/CSharp/Daemon/Stages/Dispatcher/UnityElementProblemAnalyzer.cs
```diff
@@ -19,9 +18,15 @@ protected UnityElementProblemAnalyzer([NotNull] UnityApi unityApi)
 
         protected sealed override void Run(T element, ElementProblemAnalyzerData data, IHighlightingConsumer consumer)
         {
+            // TODO: Check other process kinds. Should we run for everything?
```
citizenmatt, line 21: @krasnotsvetov can you verify what kinds we should be ignoring here? We obviously want to work for visible document and SWEA. These changes require `OTHER` to quickly find highlightings to fix, but what about the other values?

krasnotsvetov: I am not sure, that we should ignore `INCREMENTAL_SOLUTION_ANALYSIS`, but ignoring GLOBAL_WARNINGS is ok. Only performance critical inspections (and burst in future) will use global stage

PR #1653, opened 2020-05-07
WIP: Fix autosave and refresh on shared projects

nicklyu:


citizenmatt:
Can we make this some kind of extension point instead of overriding a platform component? I always get nervous when a plugin does that - what if two plugins try to do it?

nicklyu:
Hm, yep, I think you are right, I'll introduce API in rider and then update this PR

PR #1656, opened 2020-05-11
Suppress sync shared file through disk

nicklyu:


resharper/resharper-unity/src/Rider/RiderUnitySharedFilesSavingSuppressor.cs
```diff
@@ -0,0 +1,46 @@
...
+        
+        
+        public bool ShouldSuppress(IDocument document, bool forceSaveOpenDocuments)
+        {
+            if (!forceSaveOpenDocuments) return false;
+            
+            var projectFile = myDocumentToProjectFileMappingStorage.TryGetProjectFile(document);
+            var isUnitySharedProjectFile = projectFile != null 
+                                           && projectFile.IsShared() 
+                                           && projectFile.GetProject().IsUnityProject();
```
van800, line 36: I suggest to use  JetBrains.ReSharper.Plugins.Unity.ProjectModel.UnitySolutionTracker.IsUnityGeneratedProject

van800:
What would happen with redcode in different contexts.
I mean redcode fixed in one project, would not be fixed in other one.

nicklyu:
@van800 I could not reproduce this by switching contexts or whatever.

Issue #1272, opened 2019-08-07
Pause the game instead of freezing the whole Unity Editor when rider hits breakpoint


mnarimani:
I don't know the reasoning behind freezing the whole unity. But it's counter productive for me. 
When Rider hits the breakpoint, I cannot use the inspector to see the state of different objects. Which is very important in debugging.

My suggestion is that Rider should pause the game when it hits a breakpoint. This way I can still use the editor to evaluate the situation.

SugoiDev:
I don't know if that's trivial to do because it's the same process. Moreover, you might be debugging editor code so you need to halt the editor to actually do any debugging/stepping.

I use it a lot when creating custom editors and inspectors, and when debugging Unity engine issues.

mnarimani:
I think the correct way to solve this is to provide an option in `Languages & Frameworks/Unity Engine` section in settings to opt-in or opt-out this feature. 
Something like a checkbox that says `Pause the thread when a breakpoint is hit`. When the checkbox is checked, Rider should freeze the unity editor completely and when it's not checked, Rider should pause the game.

van800:
This is not technically possible. 
Unfortunately debugger protocol doesn't allow to execute code like `EditorApplication.isPaused = true`.
You may only add it to your scripts  manually.

cc: @joncham

mnarimani:
It's unfortunate that this is not possible.
Can Rider communicate to Rider plugin dll inside unity and call `EditorApplication.isPaused = true` through the plugin?

van800:
When execution is stopped on brakepoint, we can't Pause.
If we just send async Pause signal without stopping the execution, it will be only applied later in future.

mnarimani:
Alright. I will close this issue as it's not technically possible to pause the editor.

PR #1661, opened 2020-05-14
Add pausepoints

citizenmatt:
This PR adds "pausepoints", a new type of debugger breakpoint that does not suspend execution of the code, but instead pauses the Unity editor at the end of the current frame. The allows for setting a pause condition in code (e.g. hit a button, reach a certain health value or any just go down a certain code path) and then using the Unity editor to inspect the current state of the game.

To use, create a breakpoint as normal, then right click and select "Convert to Unity pausepoint".


citizenmatt:
Fixes #1272

Issue #1584, opened 2020-03-23
.asmdef / root namespace


sebastian-heinz:
Hello,

It would be great if there was a way to specify the root namespace for rider/unity projects.
I raised this in the unity forum and got redirected to this issue tracker.
Post: https://forum.unity.com/threads/rider-asmdef-namespace.852121/#post-5619076

I think a great way to solve this would be to be able to specify it inside the .asmdef file.
There us already a way to specify a package name, but using this would restrict the root namespace to the package name, not sure if this would be desired or not.

Having an additional field might require coordination with unity team.

I am unsure if VisualStudio has similiar issues or how it is handled, since .csproj files are generated on the fly it is not really possible to provide this info to the package.

Here is my .asmdef for a local package
![Capture](https://user-images.githubusercontent.com/7891178/77308682-529e7780-6d36-11ea-90aa-459a97da294e.PNG)

Here is the warning about my namespace:
![Capture](https://user-images.githubusercontent.com/7891178/77308519-09e6be80-6d36-11ea-9850-42ffab1edac4.PNG)

Maybe there are other ways to solve this, would be happy if there is a solution or maybe I am missing something that it is already possible and I am doing something wrong.

Thank you

van800:
Related to https://youtrack.jetbrains.com/issue/RIDER-36546

Strepto:
A workaround to the issue with namespaces in the Packages/ folder could be handled the same way as with namespaces in the Assets folder, strip away "Packages/" when suggesting namespace. This is not the best solution, and I would like to handle namespaces within assembly-definitions, but it helps with avoiding the basic namespace clutter.

![image](https://user-images.githubusercontent.com/3185998/77823583-737f1800-70fc-11ea-8984-7db9864f6433.png)

van800:
As a workaround you may switch from UnityExplorer to SolutionExplorer. Right click on packages folder would allow to uncheck Namespace Provider for it.
![image](https://user-images.githubusercontent.com/1482681/77996745-7c573000-732e-11ea-8f74-f2498265dc3b.png)

citizenmatt:
See also [RIDER-36546](https://youtrack.jetbrains.com/issue/RIDER-36546)

Nirlah:
Hey @citizenmatt,
Now that Unity 2020.2 added support for `rootNamespace` in .Asmdef, is there a plan to support it in Rider?

van800:
[...]

PR #1677, opened 2020-05-28
Disable namespace provider for Packages and subfolder

### LINKED ISSUE #1161: Consider stopping {package_root}/Scripts acting as a namespace provider
Rider (and ReSharper) currently prevents the `Assets` and `Assets/Scripts` folders acting as "namespace provider", so Rider won't suggest adding `Assets` or `Assets.Scripts` as part of the namespace.

Rider doesn't do anything similar for packages. So if a package root folder has a `Scripts` folder, Rider will suggest adding `Scripts` into the namespace.

To fix this, when loading each project, look for a `.asmdef` file - the location of this file marks the root folder of the package's project (a package can hold multiple projects). We can then use this path to stop `Scripts` being flagged as a namespace provider.

Big question is: is this a scenario likely to be encountered by a large number of users? Will most users set up a folder structure with `Scripts` in the root folder of a package project?

citizenmatt:
Add new rules to disable the "namespace provider" for the following folders. This will stop Rider/ReSharper from suggesting that folder as part of the namespace name. Once this PR is applied, the following folders will no longer be included in namespace name suggestions:

* `Assets`
* `Assets/Scripts`
* `Packages/<package-root>` - any top level folder under `Packages`
* `Packages/<package-root>/**` - any non-project folders. I.e. any folder under a package root that does not have `package.json`
* `Packages/<package-root>/**/Runtime`
* `Packages/<package-root>/**/Scripts`
* `Packages/<package-root>/**/Runtime/Scripts`
* `Packages/<package-root>/**/Scripts/Runtime`
* `Library/PackageCache/<package-root>`
* `Library/PackageCache/<package-root>/**` - any non-project folders. I.e. any folder under a package root that does not have `package.json`
* `Library/PackageCache/<package-root>/**/Runtime`
* `Library/PackageCache/<package-root>/**/Scripts`
* `Library/PackageCache/<package-root>/**/Runtime/Scripts`
* `Library/PackageCache/<package-root>/**/Scripts/Runtime`

This means that code in packages do not get the package name suggested, nor `Scripts`, `Runtime`, or a combination of the two. The patterns will also exclude any folder at the root of a package folder that does not contain `package.json`. This matches the use case of a git repo added to the `Packages` folder that does not have a package in the root folder of the repo (the package must be manually added via a `file:` reference in `Packages/manifest.json`)

The `Library/PackageCache` patterns stop ReSharper/Rider prompting if the user generates projects for referenced packages (which is arguably a bad idea, as these files should not be user editable, and are not in source control).

This does not fix all incorrect warnings for mismatched namespaces, as there are no widely accepted standards or conventions for namespaces or folder structure, but it does remove a number of poor suggestions. 

It is not currently possible to set a root namespace for packages (although this will be fixed in Unity 2020.2), so it is hard to get correct suggestions. The current recommendation is that the root folder of the assembly (not the package) has the same name as the root namespace. E.g. `Packages/com.unity.collections@0.0.9-preview.9/Unity.Collections/UnityCollections.asmdef` would have a root namespace of `Unity.Collections`. These changes support this scheme.

Fixes #1161 and addresses #1584 and [RIDER-36546](https://youtrack.jetbrains.com/issue/RIDER-36546)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ Since 2018.1, the version numbers and release cycle match Rider's versions and r
 ### Added
 
 - Use `Range` attribute to provide hints to integer dataflow analysis ([#1673](https://github.com/JetBrains/resharper-unity/pull/1673))
-- Remove `Packages` and any subfolder from namespace suggestions ([#1677](https://github.com/JetBrains/resharper-unity/pull/1677))
+- Improve namespace suggestions for packages ([#1161](https://github.com/JetBrains/resharper-unity/issues/1161), [RIDER-36546](https://youtrack.jetbrains.com/issue/RIDER-36546), [#1677](https://github.com/JetBrains/resharper-unity/pull/1677))
 - Rider: Add "pausepoints" a type of breakpoint that doesn't suspend code execution, but pauses the Unity editor ([#1272](https://github.com/JetBrains/resharper-unity/issues/1272), [#1661](https://github.com/JetBrains/resharper-unity/pull/1661))
 - Rider: Add sample text for "Unity", "ShaderLab" and "Cg/HLSL" Colour Scheme options pages ([#1667](https://github.com/JetBrains/resharper-unity/pull/1667))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Since 2018.1, the version numbers and release cycle match Rider's versions and r
 ### Added
 
 - Use `Range` attribute to provide hints to integer dataflow analysis ([#1673](https://github.com/JetBrains/resharper-unity/pull/1673))
+- Remove `Packages` and any subfolder from namespace suggestions ([#1677](https://github.com/JetBrains/resharper-unity/pull/1677))
 - Rider: Add "pausepoints" a type of breakpoint that doesn't suspend code execution, but pauses the Unity editor ([#1272](https://github.com/JetBrains/resharper-unity/issues/1272), [#1661](https://github.com/JetBrains/resharper-unity/pull/1661))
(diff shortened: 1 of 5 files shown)


van800:
If we remove Packages and its subfolder, then here preferred namespace would be just `Editor`
image

citizenmatt:
There's no agreed standard for namespaces or folder layouts, which makes this tricky. We disable it for `Assets`, so it makes sense that we disable it for `Packages` too. The sub folder is the name of the package, but not necessarily the name of the assembly/project, especially if the package includes multiple asmdef files, such as for tests.

It doesn't work for this example, but it does work for the scenario of taking one of Unity's own packages and copying it from `Library/PackageCache` to the `Packages` folder. E.g. the Unity Collections package has a root folder of `Packages/com.unity.collections`, but the main assembly is in `Packages/com.unity.collections/Unity.Collections`. The namespace in the code is `Unity.Collections`. The same happens with `Packages/com.unity.mathematics/Unity.Mathematics`. 

On the other hand, the assembly defined in `Packages/com.unity.ads/Runtime/Advertisement` has a namespace of `UnityEngine.Advertisements`, which we couldn't derive from the folder names.

I don't think we can get this right 100% of the time, but we can set up a good default, and I think removing the folder location and the package name is a good start. If the name of the root folder of the actual assembly is the required root namespace (`Packages/com.unity.mathematics/Unity.Mathematics`) then it works.

citizenmatt:
And that's made me think that if we generate projects for all packages, we should take a look at what happens for packages in `Library/PackageCache/<package_name>` too. Perhaps this also needs to be excluded.

van800:
You were right, that's a special case.

image

citizenmatt:
I've excluded `Library/PackageCache/<pacakge-name@version>` now as well. Looking at an example project with a load of referenced packages, I'm tempted to remove `Packages/<package_name>/Runtime`, too. I've never seen it in a namespace, and it seems to be there to mirror the `Editor` folder. What do you think?

syscrusher1709:
Thanks for this set of changes; I think this is an improvement.

PR #1703, opened 2020-06-18
Hide .meta files from navigation bar

citizenmatt:
This PR will:

* Unify the "show all files" setting between Unity Explorer and Solution Explorer
* Filter out `.meta` files from the navigation bar, as they introduce too much noise. They can be shown again with the "show all files" setting.

Fixes [RIDER-28425](https://youtrack.jetbrains.com/issue/RIDER-28425)


clearskystudio:
This still doesn't work in 2024.2.5!

MToolmaker:
@clearskystudio Hello! Please track this problem in Rider tracker:
https://youtrack.jetbrains.com/issue/RIDER-118315/Hide-.meta-file-from-navbar
Thank you!

PR #1730, opened 2020-07-02
Improve "Attach to Unity Process" dialog

citizenmatt:
This PR improves the UX of the Attach to Unity Process dialog, and fixes a few bugs in player discovery

* Editors and players are now grouped by project, when available (grouped by "Unknown project" when not)
* Editor helper processes are shown as child nodes in a tree view
* Within groups, processes are listed in order of importance - editors, then editor helpers, local/standalone players, remote players
* Local processes, such as editors and editor helpers are refreshed once a second, so new processes are added/removed while the dialog is open
* Local process discovery now only checks if a process is called "Unity", rather than trying to match heuristics. This mimics what Unity do in their own debugger plugins for other IDEs
* Player discovery is rewritten and improved:
  * Previously, polling would read one broadcast message from a socket and sleep, while more messages are being sent, creating a backlog that is never properly processed. This could prevent all players being correctly listed. Now we read all available messages
  * Previously, polling was handled with blocking and timeouts. This has been rewritten to use `select` style semantics - only do something if the socket has data
* Unity specific run configurations are now only available in Unity projects, and are dumb aware

Screenshot 2020-06-25 at 11 39 16 am


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/UnityRunUtil.kt
```diff
@@ -29,30 +29,23 @@ object UnityRunUtil {
...
+        else ""
+
+        logger.trace("isUnityEditorProcess: $name")
+        logger.trace("Checking Unity Process, execPathName: $canonicalName")
+
+        // Based on Unity's own VS Code debugger, we simply look for "Unity" or "Unity Editor". Java's
+        // ProcessInfo#executableDisplayName is the executable name with `.exe` removed. This matches the behaviour of
+        // .NET's Process.ProcessName
+        // https://github.com/Unity-Technologies/MonoDevelop.Debugger.Soft.Unity/blob/9f116ee5d344bce5888e838a75ded418bd7852c7/UnityProcessDiscovery.cs#L155
+        return (name.equals("Unity", true) || name.equals("Unity Editor", true)
```
van800, line 47: Isn't it Unity.exe on Windows?

van800: Nevermind, I see the comment above now.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/configurations/UnityAttachToEditorFactory.kt
```diff
@@ -12,6 +13,7 @@ open class UnityAttachToEditorFactory(type: ConfigurationType)
     override fun isConfigurationSingletonByDefault() = true
     override fun getName() = "Debug Unity Editor"
     override fun getIcon() = UnityIcons.RunConfigurations.AttachAndDebug
+    override fun isApplicable(project: Project) = project.isUnityProject()
```
van800, line 16: Consider using com.jetbrains.rider.UnityProjectDiscovererKt#isUnityClassLibraryProject
One may want to attach from UnityClassLib project.

van800: Check if you can attach from our rider-unity solution.

citizenmatt: Good call. I looked if `isUnityProject` worked with generated and non-generated projects, but missed projects outside of the Unity solution structure. All this does is move the configurations into the "Other" section of the new drop down for projects where it's not appropriate, so it doesn't affect much, and you can still use them and connect from any project type. I'll add the check.

Screenshot 2020-07-02 at 2 09 39 pm

Issue #1183, opened 2019-05-14
Add custom attribute to mark method as not-frequently-called


dubois:
Similarly to #1030, can we have the ability to mark methods as not being frequently called?

I work on a project that needs to run at VR frame rates on mobile hardware, so we're comfortable grappling with performance issues. However, the structure of our code creates too many false positives for the "performance critical context" feature to be useful for us. 

A lot of our code lives in manager classes which are updated once per frame; eventually the call stack leads all the way up to a singleton's `MonoBehaviour.Update()`. All this code is getting tagged as performance-critical, but we don't consider it to be "hot".

Specifically, something like this would work for us:
```
        public static bool HasFrequentlyCalledMethodAttribute(IFunctionDeclaration functionDeclaration)
        {
            return HasSpecificAttribute(functionDeclaration, "FrequentlyCalledMethodAttribute")
+               && !HasSpecificAttribute(functionDeclaration, "NotFrequentlyCalledMethodAttribute");
        }
```
and in our code
```
public class Application : MonoBehaviour {
  [AttributeUsage(AttributeTargets.Method)]
  private class NotFrequentlyCalledMethodAttribute : Attribute {
  }

  [NotFrequentlyCalled] // only once per frame
  void Update() {
    // propagate to managers ...
  }
}

```

citizenmatt:
See also [RIDER-27955](https://youtrack.jetbrains.com/issue/RIDER-27955)

FredTuna:
The attribute would be a nice addition. I would ask for something even more specific: to be able to tag a specific method call as not expensive. Within an Update function , some lines might be known to not be called every time the Update function is called (whereas the rest of the code in the Update function will be called every frame). So marking a specific method call or section of code as not expensive would help a lot in this situation.

citizenmatt:
You can already ignore an existing method invocation with a comment. Alt+Enter on the highlight, open the "Inspection" sub menu and select "Disable once with comment".

<img width="695" alt="Screenshot 2019-05-16 at 16 16 20" src="https://user-images.githubusercontent.com/222659/57865764-189b4580-77f6-11e9-8893-ad41d66c9e99.png">

FredTuna:
Yes I saw that option. Unfortunately it doesn't propagate to the method declarations. Here is an example:

Before disabling with comment:
![image](https://user-images.githubusercontent.com/23663597/57867023-62703580-77ce-11e9-90ec-5f310610c8d1.png)

After disabling with comment:
[...]

Issue #1030, opened 2019-02-08
Add custom attribute to mark method as expensive


citizenmatt:
Add a new `JetBrains.Annotations` attribute to mark a method as expensive. Any usage of this method inside a performance critical context would be highlighted as usage of an expensive method, in the same was as `GetComponent`, `Find`, etc. are currently handled.

This would be useful for third party assemblies that indirectly call the known Unity APIs that are expensive (e.g. `Camera.main` calls `FindObjectWithTag` internally) and could be included as external annotations, or compiled directly into the code. We should add features to make it easy to create or add a reference/package to use the attribute.

Another option is to ship a public attribute in the plugin, or as a package. We can mark it as `[Conditional]` to prevent a binary dependency, or maybe `[Conditional("UNITY_EDITOR")]` so the dependency is only there for the editor, which would help if the project was being distributed on the Asset Store or as a package.

We should also think carefully about choosing a name - expensive might not be expressive enough. Maybe something like "performance sensitive"? And it would be worth co-ordinating with R# team to see if it could be shared, or just something with a Unity specific name.

citizenmatt:
See also [RIDER-22807](https://youtrack.jetbrains.com/issue/RIDER-22807)

Domonion:
Implemented, from now you can add `// ReSharper enable Unity.ExpensiveCode` above your `IFunctionDeclaration` and you would get this function marked as expensive

Domonion:
Here is an example of avaiable features:   
![java_TyLQx58N8w](https://user-images.githubusercontent.com/12771085/96574014-9317a100-12d7-11eb-8aa0-59c1cba19c11.png)

And of course, they do affect the analysis transitively.

limdingwen:
![image](https://user-images.githubusercontent.com/1744967/152645822-752719c9-b1b8-449f-a63d-08a74376745a.png)

Hi, this does not seem to work for me. JetBrains Rider 2021.3.2

PR #1744, opened 2020-07-10
Transitive analyses context actions

Domonion:
I created several context actions for `Performance Critical`, `Expensive`, and `Burst` code analyses.

`Performance Critical`, `Expensive`, and `Burst` common actions:
- Enable/Disable analysis. This places `%analysis_name%CodeAnalysis[Enable/Disable]Attribute` on method and prohibits transitively propagating context through this function. This partially solves #1030, especially [https://github.com/JetBrains/resharper-unity/issues/1183#issuecomment-493119812](this) case. 
- This context actions enabled only on the method's identifier.

`Burst` only:
- Action: places `BurstDiscardAttribute` on the method. This is specially provided attribute by `Burst Compiler` that prohibits compilation of any invocations of this method.  

3 different types of action:
- `ContextAction`: methods identifier
- `QuickFix`: BurstHighlighting
- `ConfigureHighlightingProvider`: similar to `SuppressWarning`.

I am begging for some help with the following:
1. How to name attributes? Maybe `SuppressMessageAttribute` better?
2. There are too many actions and some of them should be either grouped or removed.
3. How to supply these attributes? Unity Editor already contains an instance of `JetBrains.Annotations`.
4. What exactly should be written in `Text`/`Description` sections?


Domonion:
merge into https://github.com/JetBrains/resharper-unity/pull/1852

PR #1752, opened 2020-07-14
Fixes for shader support

krasnotsvetov:


resharper/resharper-unity/src/HlslSupport/ShaderLabInclusionContextProvider.cs
```diff
@@ -30,11 +31,14 @@ public class ShaderLabInclusionContextProvider : IInclusionContextProvider
...
             };
 
+            properties.PredefinedMacros.Add(CppPPDefineSymbol.ParsePredefinedMacro("SHADER_API_D3D11"));
             var cgIncludeFolder =
                 CgIncludeDirectoryTracker.GetCgIncludeFolderPath(cache.Solution.GetComponent<UnityVersion>());
             if (!cgIncludeFolder.IsEmpty)
             {
                 properties.IncludePaths.Add(cgIncludeFolder);
+                properties.IncludePaths.Add(cache.Solution.SolutionDirectory);
+                properties.IncludePaths.Add(cache.Solution.SolutionDirectory.Combine("Library"));
```
citizenmatt, line 41: Will adding solution directory handle includes starting with `Assets/`?

Is there anything under `Library` that can be included, or is that just for `Library/PackageCache`?

resharper/resharper-unity/src/Application/UI/Options/VisualStudio/ReSharperOptionsPage.cs
```diff
@@ -64,6 +64,11 @@ public class ReSharperOptionsPage : BeSimpleOptionsPage
             AddBoolOption((UnitySettings s) => s.IsAssetIndexingEnabled,
                 "Parse text based asset files for implicit script usages (requires re-opening solution)");
 
+            
+            AddHeader("Shaders");
+            AddBoolOption((UnitySettings s) => s.SuppressShaderErrorHighlighting,
+                "Suppress resolve errors of unqualified names in HLSL");
```
citizenmatt, line 70: Why "HLSL" here, when we're currently only checking `CGPROGRAM` and `CGINCLUDE`?

resharper/resharper-unity/src/HlslSupport/UnityGameEngineDirectiveResolver.cs
```diff
@@ -0,0 +1,115 @@
...
+            mySolution = solution;
+            mySolutionTracker = solutionTracker;
+            myReferencesTracker = referencesTracker;
+            myLogger = logger;
+        }
+        
+        public bool IsApplicable(CppInclusionContext context, string path)
+        {
+              return path.StartsWith("Packages/") &&
+                   (mySolutionTracker.IsUnityProject.HasTrueValue() || myReferencesTracker.HasUnityReference.HasTrueValue());
```
citizenmatt, line 37: Do we need to do anything to support includes relative to `Assets/` here?

resharper/resharper-unity/src/HlslSupport/UnityGameEngineDirectiveResolver.cs
```diff
@@ -0,0 +1,115 @@
...
+                                {
+                                    if (!folder.IsDirectory)
+                                        continue;
+                                        
+                                    var packageAndVersion = folder.RelativePath.Name.Split('@');
+                                    if (packageAndVersion.Length != 2)
+                                        continue;
+                                    
+                                    if (result.ContainsKey(packageAndVersion[0]))
+                                        continue;
```
citizenmatt, line 93: You could use `JetSemanticVersion` on the second element and always use the highest version.

resharper/resharper-unity/src/HlslSupport/UnityHlslCppCompilationPropertiesProvider.cs
```diff
@@ -28,6 +29,10 @@ public UnityHlslCppCompilationPropertiesProvider(UnityVersion unityVersion)
                 if (!path.IsEmpty)
                     properties.IncludePaths.Add(path);
 
+                properties.PredefinedMacros.Add(CppPPDefineSymbol.ParsePredefinedMacro("SHADER_API_D3D11"));
+                properties.IncludePaths.Add(globalCache.Solution.SolutionDirectory);
+                properties.IncludePaths.Add(globalCache.Solution.SolutionDirectory.Combine("Library"));
```
citizenmatt, line 34: Same comment as `ShaderLabInclusionContextProvider` - do we have shaders that are relative to `Library`?

resharper/resharper-unity/src/HlslSupport/ShaderLabInclusionContextProvider.cs
krasnotsvetov: It is just for package cache, we could remove PackageCache and add to search path Library/PackageCache and Packages folder, but then we will need to add unity specific logic for determing priority of that folders to cpp logic, thus I decide to add Library as seperated include source folder

resharper/resharper-unity/src/Application/UI/Options/VisualStudio/ReSharperOptionsPage.cs
krasnotsvetov: Not only CGPROGRAM and CGINCLUDE, we also support HLSLPROGRAM/HLSLINCLUDE, .hlsl files, .cg files. All that files use HLSL language

resharper/resharper-unity/src/HlslSupport/UnityHlslCppCompilationPropertiesProvider.cs
krasnotsvetov: Same answer :-)

resharper/resharper-unity/src/HlslSupport/UnityGameEngineDirectiveResolver.cs
krasnotsvetov: No, when shader is used from Assets folder, it uses real file system path related to solution folder (https://youtrack.jetbrains.com/issue/RIDER-47412)

PR #1764, opened 2020-07-20
fix drag-drop/ delete operations over multiple files in Unity Explorer

van800:


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/explorer/UnityProjectModelViewExtensions.kt
```diff
@@ -78,23 +78,28 @@ class UnityProjectModelViewExtensions(project: Project) : ProjectModelViewExtens
...
+
+        val res = mutableListOf<ProjectModelNode>()
+        res.addAll(elementsWithoutProject)
+        res.addAll(elementsWithNonPlayerProject)
+
+        // there might be elements only with Player project
+        elementsWithPlayerProject.forEach { player ->
+            if (!elementsWithNonPlayerProject.any { el ->
+                    el.getVirtualFile() == player.getVirtualFile()
+                        && (el.containingProject()!!.name + ".Player" == player.containingProject()!!.name)
```
xtmq, line 97: ignore casing?

van800: there is no possibility that is would be in different casing.

PR #1765, opened 2020-07-20
Workaround for debugger evaluation timeout

van800:
I need a repro of the lock to test this.


citizenmatt:
See [RIDER-37068](https://youtrack.jetbrains.com/issue/RIDER-37068)

PR #1779, opened 2020-07-27
#RIDER-44990: Fixing tests to match the new code lens behaviour

ElenaKorotkova:


jubi20:
Fixed and pushed directly to 202

PR #1801, opened 2020-08-18
UnitTesting fixes

van800:
Temporary based on net202, would be later moved to net203.

1. Handle tests marked as [Explicit]. Avoid running them, when not asked. (RIDER-46658)
2. Handle parameterized Fixture with parameterized tests (RIDER-48686) 

Also:
https://youtrack.jetbrains.com/issue/RIDER-46658
https://youtrack.jetbrains.com/issue/RIDER-48686


resharper/resharper-unity/src/Rider/UnitTesting/RunViaUnityEditorStrategy.cs
```diff
@@ -481,43 +475,49 @@ private Task WaitForUnityEditorConnectedAndIdle(Lifetime lifetime)
...
+
+                var testNames = elements.Select(p => p.Id.Id).ToList();
+                
+                var groups = new List<string>();
+                var categories = new List<string>();
+                var criterion = run.Launch.Criterion.Criterion;
+                if (criterion is ConjunctiveCriterion conjunctiveCriterion)
                 {
-                    result.AddRange(unitTestRun.Elements);
+                    groups.AddRange( conjunctiveCriterion.Criteria.Where(a => a is TestAncestorCriterion).SelectMany(b =>
```
estrizhok, line 501: I still recommend you not to inspect criterion yourself. Instead, use `run.Elements` properties. This property returns the list of all elements that are to be executed by this run. If you want to exclude explicit elements, you can do it based on the value of `NUnitElementBase.Explicit` property. The resulting filtered list you only need to convert it into a group filter, and that's it.
The way it is written now, you are at the very least not taking into account different rules that might be inplace. For example, `RemoveElementsWithIgnoredCategoriesRule`.

estrizhok (changes requested):

PR #1817, opened 2020-08-25
Fix include resolving which points to embedded package for 2019.4+

krasnotsvetov:


citizenmatt (approved):
LGTM. If you've got this, you might as well add support for `"local"` packages, and handle packages outside of the solution. This will have a `file:` based version, which is either fully qualified, or relative to the `Packages` folder. One thing to test is that multiple `..` segments work, e.g. `file:/Users/matt/../../../../../../../../../../whatever` is a valid path, same on Windows.

resharper/resharper-unity/src/HlslSupport/Integration/Cpp/UnityGameEngineDirectiveResolver.cs
```diff
@@ -51,85 +55,154 @@ public string TransformPath(CppInclusionContext context, string path)
...
-                            var candidates = new[] {packagesFolder, packagesCacheFolder};
-                            foreach (var candidate in candidates)
+                            if (source.Equals("embedded"))
                             {
-                                foreach (var folder in candidate.GetDirectoryEntries())
+                                var packagePath = FileSystemPath.TryParse(version.RemoveStart("file:"));
+                                if (packagePath.IsEmpty)
+                                    continue;
+
+                                if (packagePath.IsAbsolute)
```
citizenmatt, line 92: Doesn't need changing, but for embedded packages, this will always be a relative path inside `Packages`, and Unity only recognises single level folders, so this will always be a single folder name. E.g. `file:com.unity.whatever`

PR #1820, opened 2020-08-26
Add internal option for suppressing errors in render-pipelines

krasnotsvetov:


resharper/resharper-unity/src/Application/UI/Options/VisualStudio/ReSharperOptionsPage.cs
```diff
@@ -73,8 +73,11 @@ public class ReSharperOptionsPage : BeSimpleOptionsPage
             {
                 AddHeader("Internal");
 
+                AddBoolOption((UnitySettings s) => s.SuppressShaderErrorHighlightingInRenderPipelinePackages, 
+                    "Suppress resolve errors in render-pipeline packages");
+                
                 AddBoolOption((UnitySettings s) => s.EnableCgErrorHighlighting,
-                    "Parse Cg files for syntax errors (requires internal mode, and re-opening solution)");
+                    "[Deprecated] Parse GLSL files for syntax errors (requires internal mode, and re-opening solution)");
```
citizenmatt, line 80: Does this do anything any more? Shall we get rid of it?

citizenmatt (approved):

resharper/resharper-unity/src/Application/UI/Options/VisualStudio/ReSharperOptionsPage.cs
krasnotsvetov: We are using old CG for GLSL (c++ team decides to not use current c++ support for glsl, thus I save old implementation to avoid any regressions). Deprecated means that this settings affects only deprecated CG support

PR #1826, opened 2020-08-27
Support macroses which comes from directives for ShaderLab

krasnotsvetov:
We could not provide full support multi_compile/shader_target directives due to complexity problem. That PR analysis directives and extract first combination of defines, also that PR covers target directive and some built-in multi_compile directives.


resharper/resharper-unity/src/HlslSupport/Integration/Injections/InjectedHlslFileLocationTracker.cs
```diff
@@ -68,91 +69,168 @@ private IEnumerable<CppFileLocation> GetIncludesLocation(IPsiSourceFile sourceFi
...
+                    // a lot of different files could be produces from multi_compile combination
+                    // Right now, we will consider first combination.
+                    if (!firstValue.IsEmpty() && (pragmaType.Equals("multi_compile") || pragmaType.Equals("multi_compile_local") ||
+                                                  pragmaType.Equals("shader_feature_local") || pragmaType.Equals("shader_feature")))
+                    {
+                        definedMacroses[firstValue] = "1";
+                    }
+
+                    if (pragmaType.Equals("target"))
+                        shaderTarget = firstValue.Replace(".", "");
```
citizenmatt, line 122: It's not obvious that `shaderTarget` is a shader version. Perhaps replace the "25", "40" and "50" magic strings with constants that explain the name? Also, is it legal to specify both `target` and `geometry` or `hull`? What should the target version be then?

citizenmatt (changes requested):

resharper/resharper-unity/src/HlslSupport/Integration/Injections/InjectedHlslFileLocationTracker.cs
krasnotsvetov:
```
            #pragma target 3.0
            #pragma geometry geom
 
            #if SHADER_TARGET == 40
                #error GEOMETRY
            #endif

```
I got "GEOMETRY" error in Unity when compiling that shader. That means that geometry is preferred by Unity thatn target directive

krasnotsvetov: Target is maximum version then, I replaced literals by constants with names

PR #1834, opened 2020-09-04
Add version to protocol

van800:
Ability to avoid connection attempt for protocol versions with different guids.
https://youtrack.jetbrains.com/issue/RIDER-48703


citizenmatt:
Don't we already have a notification about out of sync models?

van800:
@citizenmatt that notification works, when rd is the same, but model has changed.
It doesn't work good, when rd version changes.

van800:
Your question made me think, likely writing RD version there would be even better.

citizenmatt:
So this is to protect against possible breaking changes in the RD protocol? Could that version be included with the model version, so it all happens on the wire?

van800:
There is no use to add it to model, because differen RD version may not connect at all. We need this info before trying to connect.

resharper/resharper-unity/src/Rider/UnityEditorProtocol.cs
```diff
@@ -138,6 +143,11 @@ private void CreateProtocols(FileSystemPath protocolInstancePath)
 
             myLogger.Info($"EditorPlugin protocol port {protocolInstance.Port} for Solution: {protocolInstance.SolutionName}.");
 
+            if (protocolInstance.ProtocolGuid != ProtocolCompatibility.ProtocolGuid)
+            {
+                OnOutOfSync(myComponentLifetime);
```
krasnotsvetov, line 148: Return after showing notification? Otherwise protocol will be create for incompatible rd versions

krasnotsvetov (changes requested):

resharper/resharper-unity/src/Rider/UnityEditorProtocol.cs
van800: thanks!

PR #1874, opened 2020-10-01
icons for LogView

van800:
image

### DIFF
--- a/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/toolWindow/log/UnityLogPanelToolbarBuilder.kt
+++ b/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/toolWindow/log/UnityLogPanelToolbarBuilder.kt
@@ -72,7 +72,7 @@ object UnityLogPanelToolbarBuilder {
             }
         }
 
-        fun createBeforeInit() = object : ToggleAction("Messages before last [InitializeOnLoad] execution", "",  UnityIcons.LogView.FilterBeforeRefresh) {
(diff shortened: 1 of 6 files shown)


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/toolWindow/log/UnityLogPanelToolbarBuilder.kt
```diff
@@ -64,34 +65,18 @@ object UnityLogPanelToolbarBuilder {
...
-                if (isSelected(e))
-                    e.presentation.text = "Hide messages before last Play"
-                else
-                    e.presentation.text = "Show messages before last Play"
-                super.update(e)
-            }
         }
 
-        fun createBeforeInit() = object : ToggleAction("Show/Hide messages before last [InitializeOnLoad] execution", "", null) {
+        fun createBeforeInit() = object : ToggleAction("Messages before last [InitializeOnLoad] execution", "",  UnityIcons.LogView.FilterBeforeRefresh) {
```
citizenmatt, line 75: Is there a better message than this? It's very technical. Does it mean last AppDomain reload?

van800: Yes

citizenmatt: How about "Messages since last AppDomain reload"? While it's still a bit technical, everyone knows what AppDomain reloads are because they happen so often

citizenmatt (approved):

Issue #1822, opened 2020-08-27
Add Editor callbacks to performance critical context


citizenmatt:
`OnGUI` is called every frame, so should be considered performance critical context. It is usually intended for working in the Editor, but can still affect performance.

Also follow the callbacks for `EditorApplication.update`.

PR #1875, opened 2020-10-01
Added performance critical context for Editor, EditorWindow and PropertyDrawer

JurjenBiewenga:
This partially solves the issue described in #1822 however there are still events that can be subscribed on that should also be marked as performance critical


resharper/resharper-unity/src/CSharp/Daemon/Stages/PerformanceCriticalCodeAnalysis/PerformanceCriticalCodeStageUtil.cs
```diff
@@ -166,11 +165,17 @@ public static bool IsPerformanceCriticalRootMethod(ITreeNode node)
             if (typeElement == null)
                 return false;
 
-            if (!typeElement.DerivesFromMonoBehaviour())
-                return false;
+            if (!typeElement.DerivesFromMonoBehaviour() && declaration.DeclaredElement is IClrDeclaredElement monoBehaviorCLRDeclaredElement)
+                return ourKnownHotMonoBehaviourMethods.Contains(monoBehaviorCLRDeclaredElement.ShortName);
```
citizenmatt, line 169: Are these `if` checks correct? This is looking to see if the type element does NOT derive from `MonoBehaviour` but then checks against the `MonoBehaviour` lookups. Same for all of the checks here.

citizenmatt (changes requested):
Hey. Looks good. Can you check the comment, and also add some tests, please? `PerformanceStageTest` and the files are in `resharper/resharper-unity/test/data/CSharp/Daemon/Stages/PerformanceCriticalCodeAnalysis`

citizenmatt (approved):

PR #1878, opened 2020-10-02
Chrome icons are changed to match the Rider appearance

homatyano:
Icons are changed to better match the Rider appearance. Only the chrome icons (toolbar buttons and such) should be affected by this change.


citizenmatt:
The new step icon has changed, and looks different to how it does in Unity. The previous set mirrored the Unity design, which provided a useful cue as to what the buttons do. Is it worth keeping the previous design?

Screenshot 2020-10-02 at 8 45 25 pm

homatyano:
@citizenmatt Yes sure, I will fix that.

homatyano:
It appears I did a mistake here.

PR #1881, opened 2020-10-06
Recreate protocol when Rider is closed

krasnotsvetov:


unity/EditorPlugin/Utils/RiderScriptableSingleton.cs
```diff
@@ -8,6 +8,11 @@ internal class RiderScriptableSingleton: ScriptObjectSingleton<RiderScriptableSi
   {
     [SerializeField] 
     private bool myCsprojProcessedOnce;
+    
+    [SerializeField] 
+    private bool myLastPlayModeEnabled = false;
```
van800, line 13: As I recall RiderScriptableSingleton would not work at all with Unity 2019.2+

PR #1896, opened 2020-10-18
Model refactor

citizenmatt:
This PR follows on from #1887 with more refactoring of the models.

- [x] Introduce shared types in a "library" model. This makes the "passthrough" of data from fronted to backend to Unity editor (and vice versa) cleaner and easier
- [x] Introduce `PassthroughHost` to advise the backend/Unity model and pass through to the frontend/backend model (and vice versa)
- [x] Split `UnityEditorProtocol` into `BackendUnityProtocol` and `BackendUnityHost`. The protocol class is responsible for creating and recreating the protocol connection while the host class is responsible for subscribing to and exposing the model
- [x] Rename `UnityHost` to `FrontendBackendHost`
- [x] Rename some model properties
- [x] Reorder, group and add comments/documentation comments to both frontend/backend and backend/Unity models
- [x] Move protocol and host classes to `Rider/Protocol`


van800 (approved):
LGTM

PR #1902, opened 2020-10-20
Generate fake rider-dev executable on build of EditorPlugin

van800:


citizenmatt:
@van800 How does this work, and does it work for Mac? I see it generates a `.app` folder, but that contains a `.bat` file?

van800:
@citizenmatt I have checked it on Mac - you have to select that `.app` folder as External Editor in Unity.

citizenmatt:
Awesome. This is going to make life a whole lot easier! Thanks!

PR #1904, opened 2020-10-21
Change EventFunctions completion

nicklyu:
Fixing https://youtrack.jetbrains.com/issue/RIDER-52404

Also changed a bit UnityEventFunctionRule. Now if we retype method name we see only suitable suggestions.


citizenmatt (approved):
LGTM!

PR #1916, opened 2020-11-02
Fix DEXP-554913

krasnotsvetov:


van800 (approved):

citizenmatt:
I thought the protocol guid was for the version of the protocol itself, so we should only update it if the serialisation format/wire protocol changes, rather than the model changing. The model has its own serialisation hash that we listen to for out of date models, doesn't it?

van800:
For now we have only manual mechanism - when guids are different - avoid connection attempt.

PR #1917, opened 2020-11-09
Call graph related stuff

Domonion:
1. Additional CA and QF tests.
2. Local stage considers comment roots.
3. Little test fixes, optimization, and refactoring.
4. [RIDER-53010](https://youtrack.jetbrains.com/issue/RIDER-53010) and tests


resharper/resharper-unity/src/CSharp/Daemon/Stages/BurstCodeAnalysis/BurstCodeAnalysisUtil.cs
```diff
@@ -47,6 +47,15 @@ public static bool IsBurstPermittedType([CanBeNull] IType type)
             }
         }
 
+        public static bool IsQualifierOpenType(IInvocationExpression invocationExpression)
+        {
+            var result =
+                ((invocationExpression.InvokedExpression as IReferenceExpression)?.QualifierExpression as
```
krasnotsvetov, line 53: please split to several lines with variable declartions, it is hard to read that code. Why are you checking qualifier only for IReferenceExpression?

krasnotsvetov (approved):

PR #1920, opened 2020-11-13
Fix RIDER-52498

van800:
When Unity is running we attach.
When Unity is not running:
 - we start debugger, listen for incoming connections
 - we start Unity, Unity connects to debugger 

https://youtrack.jetbrains.com/issue/RIDER-52498


citizenmatt:
Why do we need a new run config for this? Why don't we just launch Unity with `-debugCodeOptimization` when it's not running and someone clicks Attach?

van800:
> Why do we need a new run config for this? Why don't we just launch Unity with `-debugCodeOptimization` when it's not running and someone clicks Attach?

Thank you for the great suggestion, now we do it all in one run-configuration.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/util/UnityUtils.kt
```diff
@@ -14,20 +15,30 @@ fun addPlayModeArguments(args : MutableList<String>) {
...
-        args.addAll(originArgs)
+    if (riderPath != null) {
+        this.addAll(mutableListOf("-riderPath", riderPath))
     }
-    return args
+    return this
+}
+
+fun MutableList<String>.withDebugCodeOptimization() : MutableList<String> {
+    this.add("-debugCodeOptimization")
```
citizenmatt, line 33: Can we get a comment for what this does please?

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/configurations/unityExe/UnityExeConfigurationType.kt
```diff
@@ -1,11 +1,10 @@
 package com.jetbrains.rider.plugins.unity.run.configurations.unityExe
 
 import com.intellij.execution.configurations.ConfigurationTypeBase
-import com.intellij.util.IconUtil
 import icons.UnityIcons
 
-class UnityExeConfigurationType : ConfigurationTypeBase("RunUnityExe", "Standalone Player", // "RunUnityExe" preserved for compatibility
-    "Unity Standalone Player configuration", UnityIcons.RunConfigurations.UnityExe) {
+class UnityExeConfigurationType : ConfigurationTypeBase("RunUnityExe", "Unity Executable", // "RunUnityExe" preserved for compatibility
+    "Either UnityEditor or Unity Standalone Player", UnityIcons.RunConfigurations.UnityExe) {
```
citizenmatt, line 7: I would be tempted to leave this as it was. We do use it internally for launching the editor if it's not already running, but we don't intend the user to use it for that purpose. As far as the user is concerned, they use the "Attach to Editor" config and that's all - they should never need to manually add a run config that launches the editor. But it is intended for use with standalone players.

van800:
done

van800:
Matt approved this.

citizenmatt (approved):

Issue #1895, opened 2020-10-17
Don't show types derived from SystemBase as unused


HugoNijmek:
Types derived from Unity.Entities.SystemBase are implicitly instantiated by Unity; it would be great not to get an unused warning for them.

citizenmatt:
See also [RIDER-46106](https://youtrack.jetbrains.com/issue/RIDER-46106)

Issue #1885, opened 2020-10-08
Incorrect redundant serialise field warning for field of the generic type of the class


citizenmatt:
```
public class Foo<T> : MonoBehaviour
{
   public T SerialisedField;
}
```

We should not show a warning that `SerialisedField` will not be serialised. We don't have enough information to tell if the generic type `T` matches the rules for Unity's serialisation.

Issue #1788, opened 2020-08-03
Custom annotation provider for Range attribute not working


citizenmatt:
The custom annotation provider recognises Unity's `[Range]` and `[Min]` attributes and applies the `[ValueRange]` annotation so that [integer data flow analysis](https://blog.jetbrains.com/dotnet/2020/04/16/resharper-ultimate-2020-1/) knows the range of these serialised fields.

This isn't working - no warnings are shown for e.g. `if` statements that check a value known to be outside of the range. This appears to be because the `[ValueRange]` attribute doesn't resolve correctly in a Unity project - the `JetBrains.Annotations.ValueRangeAttribute` type doesn't exist in Unity's copy of the annotations, while adding an implementation to the project fixes everything.

The implementation resolves against the "External annotations" PSI module, which should contain a copy of JetBrains.Annotations.dll. I thought this was enough to make the annotation resolve correctly, and thought it was working like this during testing. But testing is manual only, because external annotations are not loaded during automated testing. It might also have worked during manual testing because the editor plugin used to include an internalised version of the attributes, but have recently been removed (see #1762).

Issue #1776, opened 2020-07-24
Dictionary fields are treated as serialisable fields


citizenmatt:
`Dictionary<,>` is marked as `[Serializable]`, and since Unity 2020.1 allows generic types to be serialised, Rider treats it as a serialised field. This is incorrect - Unity 2020.1 does not allow `Dictionary<,>` to be serialised. Presumably, there is some special case handling, [similar to `System.Version`](https://github.com/JetBrains/resharper-unity/blob/5f575af7e6cd5f2b5fe05d268bb095e18421c38a/resharper/resharper-unity/src/UnityApi.cs#L122-L125).

Are there other types that should be excluded from serialisation?

citizenmatt:
See the logic in [UnitySerializationLogic.cs](https://github.com/Unity-Technologies/UnityCsReference/blob/master/Tools/Unity.SerializationLogic/UnitySerializationLogic.cs), it's more complex than the [documented rules](https://docs.unity3d.com/Manual/script-Serialization.html).

Also note that `IsNonSerialized` has a blanket exclusion for any type beginning `System.`, which explains why `System.Version` isn't serialised. There is an explicit exclusion for `Dictionary<,>`. Note that other rules seem to overlap.

PR #1923, opened 2020-11-18
Bunch of smaller fixes

### LINKED ISSUE #1785: Ensure "Use source for symbol files" is true for Unity projects
Read only packages are compiled by Unity and added as references to generated projects. As long as Unity generates PDB files, and the "Use sources from symbol files" option is checked, Rider will be able to navigate from an indexed compiled symbol to the source file. If this option is unchecked, we get decompiled source instead. 

We should check that the "Use sources from symbol files" option is checked for Unity projects. I think it would be ok to enable it automatically - it's set to true in the default settings, and makes sense to enforce for Unity projects. The alternative would be to show a notification to re-enable the setting.

citizenmatt:
- [x] Add automatic settings to ensure that navigation to source from PDB is enabled. Ensures navigation in packages goes to source files, not decompiled code. Fixes #1785 
- [x] Add automatic settings to always show implicit conversion hints, e.g. `Vector3` implicitly converted to `Vector2` loses information. These hints are push-to-hint by default
- [x] Fix some annotations - #145, #1619 
- [x] Remove checks for ECS injected fields, which are no longer used
- [x] Mark ComponentSystemBase and derived types in use - [RIDER-46106](https://youtrack.jetbrains.com/issue/RIDER-46106), #1895
- [x] Fix `[Range]` attribute analysis - #1788 
- [x] Fix presentation mode font size handling for Unity log tool window
- [x] Fix incorrect redundant `[SerializeField]` attributes on fields of type `T` - #1885 
- [x] Fix fields with `[SerializeReference]` not being recognised as serialised field - #1340 
- [x] Fix `Dictionary<TKey, TValue>` being treated as a serialisable field - #1776 
- [x] Fix incorrect redundant serialisation warnings for types such as `Vector2Int`, `BoundsInt`, etc. - see [RIDER-18729](https://youtrack.jetbrains.com/issue/RIDER-18729)
- [x] Fix `sfield` and `sprop` live templates missing in `[Serializable]` types - #1662 
- [x] Fix `.asmdef` file template missing under `Packages` folders - #780


resharper/resharper-unity/src/CSharp/Daemon/UsageChecking/UsageInspectionsSuppressor.cs
```diff
@@ -74,14 +70,15 @@ public bool SuppressUsageInspectionsOnElement(IDeclaredElement element, out Impl
                         return false;
                     }
 
-                    if (IsEventHandler(unityApi, method) || IsRequiredSignatureMethod(method))
+                    if (IsEventHandler(unityApi, method) || IsRequiredSignatureMethod(method) ||
+                        IsImplicitlyUsedInterfaceMethod(method))
```
krasnotsvetov, line 74: Is it impossible to have implicitly used interface method which is not event function?

resharper/resharper-unity/src/CSharp/Feature/Services/TypeConversionHints/CSharpTypeConversionHintsSettingsProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+        private readonly ISettingsSchema mySettingsSchema;
+        private readonly ILogger myLogger;
+
+        public CSharpTypeConversionHintsSettingsProvider(ISettingsSchema settingsSchema, ILogger logger)
+        {
+            mySettingsSchema = settingsSchema;
+            myLogger = logger;
+        }
+
+        public void InitialiseSolutionSettings(ISettingsStorageMountPoint mountPoint)
```
krasnotsvetov, line 23: I am not sure, that we should change default Rider behaviour for hints, at least in latest EAP

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/toolWindow/log/UnityLogPanelView.kt
```diff
@@ -163,7 +166,7 @@ class UnityLogPanelView(lifetime: Lifetime, project: Project, private val logMod
     @Suppress("SpellCheckingInspection")
     private val listPanel = JPanel(MigLayout("ins 0, gap 0, flowy, novisualpadding, fill", "", "[][min!]")).apply {
         add(JBScrollPane(eventList).apply { horizontalScrollBarPolicy = ScrollPaneConstants.HORIZONTAL_SCROLLBAR_NEVER }, "grow, wmin 0")
-        add(searchTextField, "growx")
+        add(searchTextField, "grow")
```
krasnotsvetov, line 169: cc @van800

resharper/resharper-unity/src/CSharp/Psi/CodeAnnotations/CustomCodeAnnotationProvider.cs
```diff
@@ -183,22 +174,123 @@ public ICollection<IAttributeInstance> GetSpecialAttributeInstances(IClrDeclared
...
+                    {
+                        if (myCtorArguments.Length == 0)
+                        {
+                            if (constructor.IsDefault) return constructor;
+
+                            continue;
+                        }
+
+                        var parameters = constructor.Parameters;
+                        if (parameters.Count == myCtorArguments.Length)
```
krasnotsvetov, line 261: that approach will not work for attributes with default parameters or if attribute arguments will be passed by name(order could be changed)

krasnotsvetov: I am not sure is it relevant to usage of your class or not

krasnotsvetov (approved):

resharper/resharper-unity/src/CSharp/Feature/Services/LiveTemplates/Scope/UnityProjectVersionScopeProvider.cs
```diff
@@ -20,28 +19,22 @@ public IEnumerable<ITemplateScopePoint> ProvideScopePoints(TemplateAcceptanceCon
                 yield break;
 
             var project = context.GetProject();
-            if (project != null && !project.IsUnityProject())
-                yield break;
+            var version = project != null
+                ? context.Solution.GetComponent<UnityVersion>().GetActualVersion(project)
+                : context.Solution.GetComponent<UnityVersion>().ActualVersionForSolution.Value;
```
van800, line 24: I would rather check ActualVersionForSolution first.

            var version = project != null
                ? context.Solution.GetComponent<UnityVersion>().ActualVersionForSolution.Value;
                : context.Solution.GetComponent<UnityVersion>().GetActualVersion(project)

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/toolWindow/log/UnityLogPanelView.kt
van800: I am not an expert in that UI tweaks, but I tried the branch and this change doesn't brake the layout of UnityLogView.

@citizenmatt what this change does?

citizenmatt: It means the search field will grow both in x and y. So it was stretching to fit horizontally, but didn't grow vertically, so if there was a larger font, the height would be wrong.

resharper/resharper-unity/src/CSharp/Daemon/UsageChecking/UsageInspectionsSuppressor.cs
citizenmatt: Yes. I don't think there was any reason to order the `IsImplicitlyUsedInterfaceMethod` before this check.

resharper/resharper-unity/src/CSharp/Psi/CodeAnnotations/CustomCodeAnnotationProvider.cs
citizenmatt: This doesn't affect usage. The attribute instance I'm creating is simple, only intended to be used to create a small number of annotation attributes and uses an array of values to find a constructor, based on correct order and type. So it's not intended to be used with default or named parameters.

resharper/resharper-unity/src/CSharp/Feature/Services/LiveTemplates/Scope/UnityProjectVersionScopeProvider.cs
citizenmatt: I think it's fine for now. We're using `GetActualVersion(IProject)` in lots of places, including `UnityApi` for completion, usage suppression and so on. I'm up for reviewing `UnityVersion` usage in 211 though - it might be worth removing the `IProject` API completely.

van800 (approved):

resharper/resharper-unity/src/CSharp/Feature/Services/TypeConversionHints/CSharpTypeConversionHintsSettingsProvider.cs
citizenmatt: This is the change I'm least sure of. I would much rather Rider showed implicit conversion hints by default, but it doesn't - they're hidden. 

My argument for doing it is that implicit type conversion hints are massively useful for highlighting issues such as possible data loss converting `Vector2` to `Vector3` and vice versa. These currently happen silently, and could easily be bugs, but we don't know intent enough to be able to show an inspection highlight. Showing a conversion hint is a great way to show this.

The setting can be changed and overridden by the user, so it can still be disabled.

PR #1945, opened 2020-11-30
Ensure we have a context when looking up types

citizenmatt:
Fixes [DEXP-564196](https://youtrack.jetbrains.com/issue/DEXP-564196)


van800 (approved):
lgtm

Issue #1946, opened 2020-11-30
Refactor Unity version discovery


citizenmatt:
The process for discovering the version of Unity used by the solution could do with being updated:

* `ActualVersionForSolution` is only ever set if the solution exists inside a Unity project folder structure, and can be null for a Unity class library with a reference to UnityEngine.dll that is not in an application install folder
* The APIs for `GetActualVersion(project)` and `ActualVersionForSolution` are inconsistent
* Updating the `ActualVersionForSolution` is only set with `AdviseOnce`, which will not catch changes to the folder structure

PR #1950, opened 2020-12-01
Refactor Unity version discovery

van800:
https://github.com/JetBrains/resharper-unity/issues/1946


resharper/resharper-unity/src/UnityVersion.cs
```diff
@@ -40,37 +42,42 @@ public class UnityVersion
...
             myUnityProjectFileCache = unityProjectFileCache;
             mySolution = solution;
+            myFileSystemTracker = fileSystemTracker;
+            
+            // SolutionDirectory isn't absolute in tests, and will throw if used with FileSystemTracker
+            mySolutionDirectory = solution.SolutionDirectory;
+            if (!mySolutionDirectory.IsAbsolute)
+                mySolutionDirectory = solution.SolutionDirectory.ToAbsolutePath(FileSystemUtil.GetCurrentDirectory());
+            
+            unitySolutionTracker.IsRelatedToUnity.WhenTrue(lifetime, SetActualVersionForSolution);
```
citizenmatt, line 52: This will fire when we have the Unity folder structure and/or Unity references. If we only have a Unity class library, we'll have references and no folder structure, but `SetActualVersionForSolution` only uses the folder structure. We'll end up setting file system watchers on incorrect paths.

I think we should use `UnitySolutionTracker.IsUnityProjectFolder.WhenTrue` for setting up the file system watchers. 

We need to make sure we call `SetActualVersionForSolution` for both Unity projects and Unity class library projects. I think we could remove `IsRelatedToUnity` and just call `SetActualVersionForSolution` in the `IsUnityProjectFolder` handler and set up another handler for `UnityReferencesTracker.HasUnityReference.WhenTrue`.

citizenmatt (changes requested):

resharper/resharper-unity/src/UnityVersion.cs
```diff
@@ -288,5 +295,10 @@ public static Version GetVersionByAppPath(FileSystemPath appPath)
             });
             return version;
         }
+
+        public void InvalidateActualVersionForSolution()
```
citizenmatt, line 299: Might be better to call it `UpdateActualVersionForSolution()` - "invalidate" suggests clearing it, but not setting it again.

citizenmatt (approved):

PR #1952, opened 2020-12-02
update to sdk 211

van800:


van800:
This was occasionally merged to net203, however this is reverted by force push now.

PR #1969, opened 2020-12-16
Net211 mf plugin build

mfilippov:


TestResult.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8" standalone="no"?>
```
van800, line 1: Remove this file?

van800 (changes requested):

rider/backend.gradle
```diff
@@ -20,6 +20,13 @@ task buildReSharperHostPlugin(type: DotNetBuildTask) {
     dependsOn prepareNuGetConfig, generateModels
 
     buildFile = backend.resharperHostPluginSolution
+    doFirst {
+        // "skipDotnet" is easier to type (and remember), but "dotNetUpToDate" is already in use
+        if (skipDotnet.equalsIgnoreCase("true") || dotNetUpToDate.equalsIgnoreCase("true")) {
+            logger.info("Skipping .NET buildReSharperHostPlugin task")
+            throw new StopExecutionException()
+        }
+    }
```
citizenmatt, line 29: How about this? I don't think we need the log, because the default is `false`, so you have to explicitly opt in (via command line parameters - `./gradlew buildPlugin -PskipDotnet=true`).

```
onlyIf { skipDotnet.equalsIgnoreCase("false") && dotNetUpToDate.equalsIgnoreCase("false") }
```

Also, I'd be happy to kill `dotNetUpToDate`. There's no real need in keeping both properties.

citizenmatt (changes requested):

unity/EditorPlugin.Tests/SlnAssetPostprocessorTest.cs
```diff
@@ -7,7 +7,7 @@ namespace JetBrains.Rider.Unity.Editor.Tests
 {
   public class SlnAssetPostprocessorTest
   {
-    [Fact]
+    // [Fact]
```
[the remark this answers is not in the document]
mfilippov: Because it doesn't work. Looks like this tests required full Unity binaries and this excluded from CI.

rider/backend.gradle
mfilippov: I want to add a log to describe skip reason.

van800 (approved):

PR #1973, opened 2020-12-17
callgraph show calls

Domonion:
4


resharper/resharper-unity/src/CSharp/Daemon/Stages/CallGraphStage/CallGraphAbstractStage.cs
```diff
@@ -79,14 +83,28 @@ public override void Execute(Action<DaemonStageResult> committer)
             committer(new DaemonStageResult(highlightingConsumer.Highlightings));
         }
 
+        private void CheckForInterrupt(IHighlightingConsumer consumer)
+        {
```
krasnotsvetov, line 87: SeldomInterruptChecker?

resharper/resharper-unity/src/CSharp/Feature/Services/CallGraph/ShowCallsBulbActionBase.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    public abstract class ShowCallsBulbActionBase : IBulbAction
+    {
+        public void Execute(ISolution solution, ITextControl textControl)
+        {
+            solution.Locks.AssertReadAccessAllowed();
+
+            var text = Text;
+
+            if (!solution.GetPsiServices().Caches.WaitForCaches(text))
```
krasnotsvetov, line 24: Check that SWA is completed at least once after start (there is property in solutionanalysisconfiguration), if swa is not completed, show tooltip

resharper/resharper-unity/src/CSharp/Daemon/Stages/CallGraphStage/CallGraphAbstractStage.cs
Domonion: CSharpDaemonProcessBase.IsProcessingFinished checks interruptFlag, seldom has it's own interrupt cookie, optimization?

PR #1996, opened 2021-01-11
Double-click on assembly in Unity Explorer should open it in AssemblyExplorer

van800:
RIDER-54873


xtmq (approved):

rider/src/main/kotlin/com/jetbrains/rider/internal/dump/DumpUnityExplorerAction.kt
```diff
@@ -0,0 +1,35 @@
+package com.jetbrains.rider.internal.dump
```
citizenmatt, line 1: I hadn't noticed where this file was placed. Can we move it to `com.jetbrains.rider.plugins.unity.actions.internal` please?

van800: Sure, will do.

PR #2003, opened 2021-01-22
Add zone marker to fix exceptions in inspectcode

citizenmatt:
When running with the Unity plugin, `inspectcode` throws an exception because a schema setting entry isn't loaded, as its owning zone (`ExternalSourcesZone`) is not enabled for the command line runner.


krasnotsvetov (approved):

rider/gradle.properties
```diff
@@ -12,7 +12,7 @@ productVersion=2020.3
 # Revision for plugin version, appended to productVersion, e.g. 2020.2.2
 # Used for published version, plus retrieving correct changelog notes
 # TODO: Should ideally come from the TC build. Manually incrementing for each release is error prone (RIDER-49929)
-maintenanceVersion=2
+maintenanceVersion=3
```
krasnotsvetov, line 15: 4?

PR #2021, opened 2021-02-11
Net211 fix run unity

evgsha:


resharper/resharper-unity/src/Rider/UnitTesting/UnityTaskRunnerHostControllerExtension.cs
```diff
@@ -0,0 +1,111 @@
...
+
+        public bool IsApplicable(IUnitTestRun run)
+        {
+            var isUnity = myUnityController.IsUnityEditorUnitTestRunStrategy(run.RunStrategy);
+            return isUnity && myAvailableProviders.ContainsKey(run.HostController.HostId);
+        }
+
+        public ClientControllerInfo GetClientControllerInfo(IUnitTestRun run, ITaskRunnerHostController next) => null;
+
+        public async Task PrepareForRun(IUnitTestRun run, ITaskRunnerHostController next)
```
estrizhok, line 60: Are you sure you don't want to call `next.PrepareForRun`, too?

resharper/resharper-unity/src/Rider/UnitTesting/UnityTaskRunnerHostControllerExtension.cs
```diff
@@ -0,0 +1,111 @@
...
+            };
+        }
+
+        public bool IsApplicable(IUnitTestRun run)
+        {
+            var isUnity = myUnityController.IsUnityEditorUnitTestRunStrategy(run.RunStrategy);
+            return isUnity && myAvailableProviders.ContainsKey(run.HostController.HostId);
+        }
+
+        public ClientControllerInfo GetClientControllerInfo(IUnitTestRun run, ITaskRunnerHostController next) => null;
```
estrizhok, line 58: if you don't need this method, can I remove it from the interface? I only added it because I saw that dotCover's variant of writing an extension was overriding `controllerInfo`. I would actually like to delete `GetClientControllerInfo` from 'ITaskRunnerHostContoller' altogether. This method is only related to a specific implementation of runners, and should not be in base interface.

estrizhok (commented):
please see inline comments

resharper/resharper-unity/src/Rider/UnitTesting/UnityTaskRunnerHostControllerExtension.cs
evgsha: Yes, I agree, but I think that you mean delete GetClientControllerInfo from ITaskRunnerHostContoller**Extension**?

estrizhok: Both :)

evgsha: Ok, I've understood

resharper/resharper-unity/src/Rider/UnitTesting/UnityTaskRunnerHostControllerExtension.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            var lifetimeDef = myLifetime.CreateNested();
+            run.PutData(ourLifetimeDefinitionKey, lifetimeDef);
+
+            lock (myStartUnitySync)
+            {
+                WrapStartUnityTask(() => PrepareForRunInternal(lifetimeDef.Lifetime, run));
+                WrapStartUnityTask(() => next.PrepareForRun(run));
+                
+                return myStartUnityTask;
```
estrizhok, line 72: I don't understand all of this task wrapping. Couldn't it just be?
```
await PrepareForRunInternal(lifetimeDef.Lifetime, run);
await next.PrepareForRun(run);
```
Or did you explicitly want to run both task (yourself and next) concurrently?
``` 
return Task.WhenAll(
  PrepareForRunInternal(lifetimeDef.Lifetime, run),
  next.PrepareForRun(run)
);
```
Also, why create nested lifetime inside this method, if it is only used it PrepareForRunInternal?
It seems unnecessary complicated

estrizhok (commented):
more comments

PR #2023, opened 2021-02-12
Shorter Fix Rider-50085 Stopping "Attach and Play" should always stop both Debug and Play

van800:


van800:
Comments from @krasnotsvetov:
1. It might be a feature not worth the fix. No one from real users asked for it.
2. If I decide to merge it, I should make it return to the play state, which was before running the debug&play configuration. Instead of plain setting mode to EditMode.

resharper/resharper-unity/src/Rider/Protocol/BackendUnityHost.cs
```diff
@@ -69,7 +69,13 @@ public class BackendUnityHost
...
                 var frontendBackendModel = frontendBackendHost.Model.NotNull("frontendBackendHost.Model != null");
                 BackendUnityModel.FlowIntoRdSafe(lifetime,
-                    backendUnityModel => backendUnityModel != null,
+                    backendUnityModel =>
+                    {
+                        var res = backendUnityModel != null;
+                        // when connection to Unity disappears, state of PlayControls in the Frontend-Backend protocol may not correspond to state in Unity
+                        if (!res) frontendBackendHost.Model?.PlayControlsInitialized.SetValue(false);
+                        return res;
+                    },
```
citizenmatt, line 78: This should be moved to the passthrough host. The backend/unity host should not have a reference to the frontend/backend as it makes it much easier to introduce circular dependencies. The passthrough host is where all other frontend <-> unity communication happens.

The backend/unity host only has a reference to the frontend host to handle the editor state polling, which should really be handled with a property and should really be part of the passthrough host (I left it here because it felt more like infrastructure than passing data around, but it's still passthrough stuff)

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/configurations/UnityAttachToEditorProfileState.kt
```diff
@@ -42,11 +45,20 @@ class UnityAttachToEditorProfileState(private val exeDebugProfileState : UnityEx
...
                         if (it == DebuggerInitializingState.Initialized) {
                             logger.info("Pass value to backend, which will push Unity to enter play mode.")
+                            var prevState = false
                             lt.bracket(opening = {
                                 // pass value to backend, which will push Unity to enter play mode.
+                                prevState = executionEnvironment.project.solution.frontendBackendModel.playControls.play.valueOrDefault(false)
                                 executionEnvironment.project.solution.frontendBackendModel.playControls.play.set(true)
                             }, terminationAction = {
-                                executionEnvironment.project.solution.frontendBackendModel.playControls.play.set(false)
+                                val project = executionEnvironment.project
```
citizenmatt, line 54: Add a comment why this necessary, i.e. if termination happens before the protocol connection is made, the action is lost

citizenmatt (changes requested):

citizenmatt (approved):

PR #2026, opened 2021-02-15
enable logging for each assembly

van800:
logs example https://buildserver.labs.intellij.net/buildConfiguration/DotNetDiv_Wave211_Rider_Plugins_Unity_BuildPlugin/107348983?buildTab=artifacts


citizenmatt (approved):
For extra context: the test framework sets up default verbose logging and saves the main log file to `resharper.log`, which will get overwritten by subsequent test assemblies. The `SetupLogging` method in `TestEnvironment.cs` will set up verbose logging earlier in the process, capturing more details, and with a name unique to the assembly.

PR #2030, opened 2021-02-17
Fix meta files not being updated

### LINKED ISSUE #2027: Renaming classes does not update .meta file
**Description:**

Renaming classes via Rider does not update .meta file, breaking all references in a project.

1. Create a class
image

2. Rename it
image

3. Watch as .cs.meta filename stays the same and does not get named by this plugin

image

Unity Support plugin version:

image

**Expected:**

Renaming a class should also rename associated .meta file, so the reference guid stays the same.

Unity settings:
image

citizenmatt:
This PR fixes two issues, mostly related to `.meta` file tracking and updating:

- [x] A solution is recognised as generated if the name is the same, but the case is different. Unity should generate the name correctly, but some other tooling might not. Also, if the case is different in the recent solutions list, Rider uses the incorrect case. If we don't recognise a solution as generated, meta files aren't updated, which is bad. Fixes #2027
- [x] Change meta file tracker to work with any solution inside a Unity project - generated or manual. If we're moving things inside `Assets` or `Packages`, we should update the `.meta` files.

### DIFF
--- a/resharper/resharper-unity/src/ProjectModel/MetaFileTracker.cs
+++ b/resharper/resharper-unity/src/ProjectModel/MetaFileTracker.cs
@@ -21,9 +21,9 @@ namespace JetBrains.ReSharper.Plugins.Unity.ProjectModel
         private readonly ILogger myLogger;
         private IProjectItem myLastAddedItem;
 
-        public MetaFileTracker(Lifetime lifetime, ChangeManager changeManager, ISolution solution, ILogger logger, 
-            ISolutionLoadTasksScheduler solutionLoadTasksScheduler,
-            UnitySolutionTracker unitySolutionTracker)
+        public MetaFileTracker(Lifetime lifetime, ChangeManager changeManager, ISolution solution, ILogger logger,
+                               ISolutionLoadTasksScheduler solutionLoadTasksScheduler,
+                               UnitySolutionTracker unitySolutionTracker)
         {
             mySolution = solution;
             myLogger = logger;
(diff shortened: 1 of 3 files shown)


resharper/resharper-unity/src/ProjectModel/MetaFileTracker.cs
```diff
@@ -21,19 +21,19 @@ public class MetaFileTracker : IChangeProvider
...
+                               UnitySolutionTracker unitySolutionTracker)
         {
             mySolution = solution;
             myLogger = logger;
 
             solutionLoadTasksScheduler.EnqueueTask(new SolutionLoadTask("AdviseForChanges", SolutionLoadTaskKinds.AfterDone,
                 () =>
                 {
-                    if (!unitySolutionTracker.IsUnityGeneratedProject.Value)
+                    if (!unitySolutionTracker.IsUnityProjectFolder.Value)
```
van800, line 34: Is this intentional? You have already changed `IsUnityGeneratedProject` to accept wrong casing.
I am affraid of custom valid solution in UnityProject folder, which may include some non-generated projects.

citizenmatt: Yeah, intentional. The casing issue is still a problem we should fix, even if we don't use it for meta files. And I think if we're changing files in the `Assets` or `Packages` folders with a custom solution, then we should handle meta files too. We know it's a Unity solution, so we should do our best not to break things.

van800: Ok, I see your point.

van800 (approved):

PR #2040, opened 2021-03-02
Fix completion when replacing entire method name

citizenmatt:
When editing the name of a function in a Unity class, if the name is being completely replaced, e.g. changing `Awake` to `Start` then the auto-completion for event functions will generate incorrect code. It will generate a new body for the event function, while the existing body is still there, resulting in something like:

```
public void Awake()
{
}

()
{
  // Existing code
}
```

### DIFF
--- a/resharper/resharper-unity/test/data/CSharp/CodeCompletion/Action/RetypeFullNameOnExistingMethod02.cs
+++ b/resharper/resharper-unity/test/data/CSharp/CodeCompletion/Action/RetypeFullNameOnExistingMethod02.cs
@@ -1,0 +1,10 @@
+// ${COMPLETE_ITEM:OnApplicationPause(bool) { ... \}}
+using UnityEngine;
+
+public class A : MonoBehaviour
+{
+  public void {caret}(bool hasFocus)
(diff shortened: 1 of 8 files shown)


resharper/resharper-unity/test/data/CSharp/CodeCompletion/Action/RetypeFullNameOnExistingMethod02.cs
```diff
@@ -0,0 +1,10 @@
+// ${COMPLETE_ITEM:OnApplicationPause(bool) { ... \}}
+using UnityEngine;
+
+public class A : MonoBehaviour
```
nicklyu, line 4: @citizenmatt May be we should use UnterminatedContext reparse in this case? Instead of `var updateMethodDeclaration = TextControlToPsi.GetElement<IMethodDeclaration>(solution, textControl);
`
It inserts __ in caret position and we can work with 
`  public void __{caret}(bool hasFocus)
  {
    // Existing code
  }`
  
  So it will parse as correct method declaration and doesn't break anything

PR #2045, opened 2021-03-12
ShowFileInUnityAction in Unity/Solution Explorer

van800:
RIDER-59405 

image

image


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/actions/ShowFileInUnityAction.kt
```diff
@@ -0,0 +1,54 @@
...
+    override fun actionPerformed(e: AnActionEvent) {
+        val project = e.project ?: return
+        val file = getFile(e) ?: return
+
+        execute(project, file)
+    }
+
+    override fun update(e: AnActionEvent) {
+        val project = e.project ?: return
+        e.presentation.isVisible = project.isUnityProjectFolder()
```
krasnotsvetov, line 27: If getFile returns false, action should not be visible

krasnotsvetov (changes requested):

PR #2050, opened 2021-03-22
OnlineHelp or Local Documentation by F1 + fixes in quick doc link to external doc

van800:
1. Quick Documentation shortcut opens Unity types and event functions in locally installed documentation or on `docs.unity3d.com`.
image
RIDER-33108 Allow to open Unity local documentation.

Need to test Re# version.

2. Links in QuickDoc popup:
Show links to dotnet/mono/unity documentation  
RIDER-55736 Tweak external documentation link for generic methods
RIDER-39068 Improve QuickDoc API to include external documentation URL
RIDER-57745 Wrong page opens in external documentation
RIDER-60297 External Doc link for parameter of event function should go to eventfunction doc instead
RIDER-55737 Allow customizing QuickDoc link to use localized Unity documentation


resharper/resharper-unity/src/Application/UI/Help/ShowUnityHelp.cs
```diff
@@ -37,25 +36,39 @@ public bool ShowHelp(string keyword, HelpSystem.HelpKind kind, string preferredP
         public double Priority => 0;
         public HelpSystem.HelpKind[] SupportedKinds => new[] {HelpSystem.HelpKind.Msdn};
 
-        private static bool IsUnityKeyword(string keyword)
+        public static bool IsUnityKeyword(string keyword)
         {
-            return keyword.StartsWith("UnityEngine.", StringComparison.OrdinalIgnoreCase)
-                   || keyword.StartsWith("UnityEditor.", StringComparison.OrdinalIgnoreCase);
+            return keyword !=null &&
```
citizenmatt, line 41: What does it mean to be called with `null`? What help is ReSharper trying to show in that case? And if you've got the chance, can you add `[CanBeNull]` to the interface?

resharper/resharper-unity/src/Application/UI/Help/ShowUnityHelp.cs
```diff
@@ -37,25 +36,39 @@ public bool ShowHelp(string keyword, HelpSystem.HelpKind kind, string preferredP
...
         {
-            return keyword.StartsWith("UnityEngine.", StringComparison.OrdinalIgnoreCase)
-                   || keyword.StartsWith("UnityEditor.", StringComparison.OrdinalIgnoreCase);
+            return keyword !=null && 
+                   (keyword.StartsWith("UnityEngine.", StringComparison.OrdinalIgnoreCase)
+                   || keyword.StartsWith("UnityEditor.", StringComparison.OrdinalIgnoreCase));
         }
 
-        private static string StripPrefix(string keyword)
+        public static string StripPrefix(string keyword)
```
citizenmatt, line 46: This shouldn't be public. It makes a huge assumption about the value of the text, so we should only call it from this context.

resharper/resharper-unity/src/CSharp/Feature/OnlineHelp/UnityCompiledElementOnlineHelpProvider.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+        
+        [Pure, CanBeNull]
+        private static string GetSearchableText(ICompiledElement compiledElement)
+        {
+            if (compiledElement is ITypeElement)
+            {
+                return compiledElement.ShortName;
+            }
+            return ShowUnityHelp.StripPrefix(compiledElement.GetSearchableText());
```
citizenmatt, line 46: Don't use `ShowUnityHelp.StripPrefix`. That method is assuming that there are 12 characters to strip, and there's no checking that assumption here. Create a method here that's appropriate for whatever `GetSearchableText` returns (and it's a good idea to check what it returns rather than make another assumption)

resharper/resharper-unity/src/CSharp/Feature/OnlineHelp/UnityCompiledElementOnlineHelpProvider.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public override int Priority => 10;
```
citizenmatt, line 49: I like to document where a priority like this puts us in relation to the default providers. Is this first, or after the default MSDN provider, etc.?

resharper/resharper-unity/src/CSharp/Feature/OnlineHelp/UnityOnlineHelpProvider.cs
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        public Uri GetUrl(IDeclaredElement element)
+        {
+            if (!IsAvailable(element)) return null;
+            var unityApi = element.GetSolution().GetComponent<UnityApi>();
+            var name = element.GetUnityEventFunctionName(unityApi);
+            
+            if (!ShowUnityHelp.IsUnityKeyword(name)) return null;
+            var keyword = ShowUnityHelp.StripPrefix(name);
```
citizenmatt, line 27: It's a bit weird that we have this implementation split across two classes - they're doing the same thing, why do we need two classes? Does ReSharper still use `IShowHelp`? Could it use `IOnlineHelpProvider` instead? Or provide a default implementation that uses `IOnlineHelpProvider`?

resharper/resharper-unity/src/CSharp/Feature/Services/QuickDoc/UnityEventFunctionQuickDocPresenter.cs
```diff
@@ -67,8 +68,12 @@ private XmlNode GetDetails(IDeclaredElement element)
...
 
             var details = CreateMemberElement(element);
-            if (!string.IsNullOrWhiteSpace(description))
-                details.CreateLeafElementWithValue("summary", description);
+            if (string.IsNullOrWhiteSpace(description)) return details;
+            details.CreateLeafElementWithValue("summary", description);
+            if (!element.GetPsiServices().Solution.HasComponent<IXmlDocLinkAppender>()) return details;
+            var uri = element.GetPsiServices().Solution.GetComponent<UnityOnlineHelpProvider>().GetUrl(element);
+            element.GetPsiServices().Solution.GetComponent<IXmlDocLinkAppender>().AppendExternalDocumentationLink(uri, element.ShortName, details);
+
```
citizenmatt, line 76: Does `AppendExternalDocumentationLink` need to be an instance method on `IXmlDocLinkAppender`? Feels like it might be simpler as a static method on something like `XmlDocHtmlUtil`

citizenmatt (changes requested):

resharper/resharper-unity/src/CSharp/Feature/Services/QuickDoc/UnityEventFunctionQuickDocPresenter.cs
van800: I use check for `IXmlDocLinkAppender` presence to add link for Rider and avoid adding it for Re#.

resharper/resharper-unity/src/CSharp/Feature/OnlineHelp/UnityOnlineHelpProvider.cs
van800: I haven't found a way unify IShowHelp and IOnlineHelpProvider.
IShowHelp is coupled with QuickDoc for Re# in VS and IOnlineHelpProvider goes together with ContextHelpActionHandler , which is used with F1 (External Documentation) in VS/Rider.

resharper/resharper-unity/src/Application/UI/Help/ShowUnityHelp.cs
```diff
@@ -76,7 +85,11 @@ private FileSystemPath GetDocumentationRoot()
         {
             var appPath = mySolutionsManager.Solution?.GetComponent<UnityVersion>().GetActualAppPathForSolution();
             var contentsPath = UnityInstallationFinder.GetApplicationContentsPath(appPath);
-            return contentsPath.Combine(@"Documentation/en");
+            var root = contentsPath.Combine("Documentation");
+            var englishRoot = root.Combine("en");
+            if (!englishRoot.ExistsDirectory)
+                return root.GetChildDirectories().FirstOrDefault(englishRoot);
+            return englishRoot;
```
citizenmatt, line 92: Oh wow. Does this mean they're installing non-English documentation now?

citizenmatt (approved):
LGTM!

resharper/resharper-unity/src/Application/UI/Help/ShowUnityHelp.cs
van800: Not really. It can be downloaded separately and manually put to the folder
https://docs.unity3d.com/Manual/OfflineDocumentation.html

PR #2063, opened 2021-04-08
Disable ShowMarkerOnStaticMethods for Unity projects

van800:
Hide run-markers, because for Unity projects this wouldn't work.
For Unity we have `UnityRunMarkerProvider` instead.

https://youtrack.jetbrains.com/issue/RIDER-55734


citizenmatt (commented):
LGTM!

PR #2076, opened 2021-04-20
Fixes for additional content roots in workspace model

citizenmatt:
This PR will:

- [x] Only add Unity specific folders to the content model for a Unity project. This is especially important for `<project-root>/Packages`
- [x] Reset the extra folders and includes/excludes in the content model stored in `indexLayout.xml` by previous versions. The upgrade flag is stored in global config, keyed with the hash of the `.idea` path

### DIFF
--- a/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/workspace/UnitySolutionManagerExtensions.kt
+++ b/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/workspace/UnitySolutionManagerExtensions.kt
@@ -9,22 +9,25 @@ import java.nio.file.Path
 class UnitySolutionManagerExtensions : SolutionManagerExtensions {
 
     companion object {
-        private const val CLEANUP_USER_CONTENT_MODEL_KEY = "user.content.model.migrated"
+        private const val CLEANUP_USER_CONTENT_MODEL_PREFIX = "unity.user.content.model.migrated-"
     }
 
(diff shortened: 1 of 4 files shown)


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/workspace/UnitySolutionManagerExtensions.kt
```diff
@@ -9,22 +9,25 @@ import java.nio.file.Path
...
+        val key = CLEANUP_USER_CONTENT_MODEL_PREFIX + dotIdeaDir.hashCode().toString(16)
+        val handledPath = PropertiesComponent.getInstance().getValue(key)
+        if (handledPath == dotIdeaDir.toString()) return emptyArray()
 
-        return super.cleanupBeforeOpen(dotIdeaDir)
+        // <projectDir>/.idea/.idea.SolutionName/.idea
+        val projectDir = dotIdeaDir.parent?.parent?.parent?.toFile()?.toVirtualFile(true) ?: return emptyArray()
+        if (!UnityProjectDiscoverer.hasUnityFileStructure(projectDir)) return emptyArray()
+
+        PropertiesComponent.getInstance().setValue(key, dotIdeaDir.toString())
```
xtmq, line 29: setValue(..., true) maybe instead of full path? just a flag

xtmq (approved):

PR #2080, opened 2021-04-28
Minor issue with attempt to Profile from UnityClassLib

van800:
Only show "Unable to automatically discover correct Unity Editor to debug" for debug. Avoid showing it for Profile.

### DIFF
--- a/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/configurations/UnityAttachToEditorRunConfiguration.kt
+++ b/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/configurations/UnityAttachToEditorRunConfiguration.kt
@@ -99,6 +99,17 @@ class UnityAttachToEditorRunConfiguration(project: Project, factory: Configurati
         configurationPerRunnerSettings: ConfigurationPerRunnerSettings?
     ) {
         if (runner is DotNetDebugRunner) {
+            // This method lets us check settings before run. If we throw an instance of RuntimeConfigurationError, the Run
+            // Configuration editor is displayed. It's called on the EDT, so there's not a lot we can do - e.g. we can't get
+            // a process list.
+
+            // If we already have a pid, that means this run configuration has been launched before, and we've successfully
+            // attached to a process. Use it again. If the pid is out of date (highly unlikely), we'll do our best to find
+            // the process again
+            if (pid != null) {
+                return
+            }
+
             // If we're a class library project that isn't in a Unity project folder, we can't guess at the correct project
             // to attach to, so throw an error and show the dialog. This value will be null until the backend has finished
             // loading. However, because we're a Unity run configuration, we can safely assume we're a Unity project, and if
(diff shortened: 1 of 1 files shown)


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/configurations/UnityAttachToEditorRunConfiguration.kt
```diff
@@ -92,6 +93,26 @@ class UnityAttachToEditorRunConfiguration(project: Project, factory: Configurati
...
+
+            val isClassLibraryProject = project.isUnityClassLibraryProject()
+            if (!project.isUnityProjectFolder() && (isClassLibraryProject == null || isClassLibraryProject)) {
+                throw RuntimeConfigurationError("Unable to automatically discover correct Unity Editor to debug")
+            }
+        }
+        super.checkRunnerSettings(runner, runnerSettings, configurationPerRunnerSettings)
+    }
+
     override fun checkSettingsBeforeRun() {
```
citizenmatt, line 116: This method could be deleted now.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/configurations/UnityAttachToEditorRunConfiguration.kt
```diff
@@ -92,6 +93,26 @@ class UnityAttachToEditorRunConfiguration(project: Project, factory: Configurati
...
+        configurationPerRunnerSettings: ConfigurationPerRunnerSettings?
+    ) {
+        if (runner is DotNetDebugRunner) {
+            // If we're a class library project that isn't in a Unity project folder, we can't guess at the correct project
+            // to attach to, so throw an error and show the dialog. This value will be null until the backend has finished
+            // loading. However, because we're a Unity run configuration, we can safely assume we're a Unity project, and if
+            // we're not inside a Unity project folder, then we can't automatically attach, so throw an error and show the
+            // dialog
+
+            val isClassLibraryProject = project.isUnityClassLibraryProject()
```
citizenmatt, line 108: This needs a check for `pid != null` first. If we've previously attached to an editor, we try to reconnect. I think this logic is still true for class libraries.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/configurations/UnityAttachToEditorRunConfiguration.kt
```diff
@@ -92,6 +93,26 @@ class UnityAttachToEditorRunConfiguration(project: Project, factory: Configurati
         return null
     }
 
+    override fun checkRunnerSettings(
```
citizenmatt, line 96: Does this method have the same result as `checkSettingsBeforeRun`? I seem to remember that `checkSettingsBeforeRun` would show the editor form, so we can choose an editor to attach to, while `checkRunnerSettings` throws an exception and shows it as a balloon notification. Have I remembered that right?

citizenmatt (changes requested):

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/run/configurations/UnityAttachToEditorRunConfiguration.kt
van800: they both show the editor form for me.

citizenmatt (approved):

PR #2082, opened 2021-05-03
Update plugin testing on Mac/Linux

citizenmatt:
Fixes a number of issues to get testing working again on Mac/Linux:

- [x] Removes some redundant hacks working around severe performance problems with the implementation of `System.IO.Packaging` on Mono. The ReSharper SDK uses a newer version of NuGet which means we no longer need to patch Mono's `OptimizedZipPackage.EnsurePackageFiles`. However, there are still some usages of NuGet 2.x that call Mono's really bad implementation of `ZipPackage.LoadParts` and that can add about 5 seconds to a 30 second test run. So this hack remains, for now.
- [x] Add an `ILanguageHlslSupportZone` zone definition which is only available if the `ILanguageCppZone` is available. Mono can't load the Cpp PSI assembly, and since it's not loaded, the `ILanguageCppZone` is not loaded or activated, and so all HLSL support code is disabled for Mac tests. An additional `[RequireHlslSupport]` attribute is available for test classes and methods that will ignore the tests if the `ILanguageHlslSupportZone` is not available.
- [x] Force the test assemblies to be compiled as x86 on Windows only. We need to specify 32 or 64 bit in order to load the correct Cpp PSI assembly, but Mac only supports 64 bit processes, and Mono doesn't support the Cpp PSI assembly. Windows stays as 32 bit, Mac/Linux use the OS default, which is 64 bit for Mac.
- [x] Disables the verbose logging set up in the `TestEnvironment` classes. This has a measurable impact on performance (it's the third `VERBOSE` file logger, and the constructor that the config uses is costly when called for every test). On my Mac, it drops a ~35 second run to ~29 seconds. It was added to capture startup logging, to help diagnose issues running tests on Mac, and isn't needed for day-to-day running. The default logs already output per-test and per-test run logs at `VERBOSE`, it just doesn't include startup logging.
- [x] Move the location of the default logs. The per-test run logs are written to `$TMPDIR/JetLogs/ReSharperTests/resharper.log`, which means that the logs are not per-assembly. The same file should be reused if the timestamp isn't older than 3 hours, but now the logs are created in `{solution-root}/resharper/build/{project}/logs` which means we get separate logs for rider-unity, rider-yaml, resharper-unity and resharper-yaml.

I haven't tested on Linux, but it should have the same behaviour as Mono on Mac.

### DIFF
--- a/resharper/resharper-unity/src/HlslSupport/ZoneMarker.cs
+++ b/resharper/resharper-unity/src/HlslSupport/ZoneMarker.cs
@@ -4,8 +4,13 @@
 namespace JetBrains.ReSharper.Plugins.Unity.HlslSupport
 {
   [ZoneMarker]
-  public class ZoneMarker :
-    IRequire<ILanguageCppZone>
+  public class ZoneMarker : IRequire<ILanguageHlslSupportZone>
+  {
+  }
+
+  // By inheriting from ILanguageCppZone, activation is also inherited
+  [ZoneDefinition]
+  public interface ILanguageHlslSupportZone : ILanguageCppZone
   {
   }
 }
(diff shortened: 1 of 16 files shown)


citizenmatt:
I think we'll need to update the TeamCity config to collect logs from the new location.

van800 (approved):
LGTM

PR #2084, opened 2021-05-06
Fix issues resolving packages

citizenmatt:
This PR will fix a couple of edge cases while resolving packages:

- [x] Fix an issue displaying a package dependency that has been resolved when the [package dependency resolution strategy](https://docs.unity3d.com/Manual/upm-manifestPrj.html#strategize) is set to `highestPatch`, `highestMinor` or `highest`. In this case, Unity can resolve a dependency to a higher version than requested in a requiring package's `package.json`. Rider doesn't have this information, so resolves the dependency against the contents of `Library/PackageCache`.
- [x] Fix an issue displaying a Git based package when the lock file is disabled. There is no hash to use to look up the package in the local package cache. Instead, we look for the package ID as a prefix in the local package cache.
- [x] Add extra checks for packages that take precedence over resolved dependencies. Previously, the code only considered embedded packages as taking precedence. As [per the docs](https://docs.unity3d.com/Manual/upm-conflicts.html), any embedded, local, git or built in packages take precedence over a requested version in a dependency.

### DIFF
--- a/resharper/resharper-unity/src/Packages/PackageData.cs
+++ b/resharper/resharper-unity/src/Packages/PackageData.cs
@@ -83,10 +83,10 @@ namespace JetBrains.ReSharper.Plugins.Unity.Packages
     public class GitDetails
     {
         [NotNull] public readonly string Url;
-        [NotNull] public readonly string Hash;
+        [CanBeNull] public readonly string Hash;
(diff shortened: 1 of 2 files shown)


van800 (approved):
LGTM!

PR #2090, opened 2021-05-17
Allow Editor templates for projects with UNITY_EDITOR define

van800:
https://jetbrains.slack.com/archives/C53QFRZ7G/p1620924775194600

When we have asmdef with Editor
"includePlatforms": [
    "Editor",
],
UNITY_EDITOR define is in the csproj

Otherwise - when asmdef either exсludes Editor or doesn't include it - UNITY_EDITOR define is not in the csproj
"excludePlatforms": [
    "Editor"
]

Assembly-CSharp and Assembly-CSharp-firstpass contain UNITY_EDITOR define, however Editor api-s compilation would fail.
If Player projects generation is enabled - those Player projects would not contain UNITY_EDITOR define.


citizenmatt (approved):
I think this is too broad - since `UNITY_EDITOR` is defined for all projects that are open in the editor, this means we'll show editor templates in runtime code for packages and assembly definitions. But I think it's better than not showing them at all for these projects.

PR #2098, opened 2021-05-24
Support debugging local UWP players + other debugger extension updates

citizenmatt:
This PR:

- [x] Adds support for debugging local UWP players. UWP processes are not allowed to accept incoming socket connections from the local machine in order to [isolate Windows Store apps](https://docs.microsoft.com/en-us/previous-versions/windows/apps/dn640582(v=win.10)). Remote players are not affected. This change will launch the `CheckNetIsolation` tool to temporarily enable loopback connections. Fixes [RIDER-62321](https://youtrack.jetbrains.com/issue/RIDER-62321)
- [x] Ignores the "break on unhandled exceptions" setting for IL2CPP players. This matches the behaviour for the Mono 4.x runtime, which ignores this setting. I'm not clear why yet. This behaviour is controlled by an option, so we can return to the original behaviour if required. This fixes [RIDER-62321](https://youtrack.jetbrains.com/issue/RIDER-62321). Ignoring this setting also works around some other issues with IL2CPP. E.g. an exception thrown from a call stack with no source at all ([RIDER-62320](https://youtrack.jetbrains.com/issue/RIDER-62320)).
- [x] Refactors debugger worker extension options to use the protocol rather than environment variables. Options are now updated live in the debugger worker process, but won't necessarily redraw until state changes (e.g. changing the "extend value rendering" option doesn't force redrawing the list of variables)
- [x] Refactors iOS USB debugging proxy creation to use proper start info handler mechanism over the protocol, rather than cheating with environment variables


van800 (approved):
LGTM

PR #2101, opened 2021-06-01
fix DEXP-589706

van800:


citizenmatt (approved):
🙀 Oh wow. This value isn't set for non-Unity projects. Silly mistake. Thanks for the fix! 🎉

PR #2106, opened 2021-06-15
Code analysis optimization

krasnotsvetov:
1. Do not use component locator when checking IsUnityProject
2. Do not recalculate attributes each time in CustomCodeAnnotationProvider
3. Cache "is unity type" caching


citizenmatt (approved):
LGTM

PR #2119, opened 2021-07-12
Prevent break on unhandled ExitGUIException

citizenmatt:
The Unity 2021.2 beta has an upgraded Mono which now reports unhandled exceptions to the debugger. The previous Mono version (~5.12ish) didn't report unhandled exceptions. This old behaviour does not appear to be deliberate or intentional, and it's not yet clear if this behaviour will be rolled back or not.

However, it has the consequence that the debugger will now frequently stop on unhandled `ExitGUIException`s. These are used as control flow exceptions, thrown in managed code and caught in native code (and therefore considered "unhandled" by Mono and Rider) in order to cancel behaviour in the GUI. The exception is thrown frequently as the user works with the Inspector and other windows, and makes using the Unity editor while debugging very annoying. This PR will ignore any unhandled `ExitGUIException`s for Unity projects.

If the behaviour is rolled back (or even back ported to older Unity versions), the feature is safe - it will only suppress unhandled exceptions with a full name of `UnityEngine.ExitGUIException`. If Unity stop reporting the exception, there will be no exception to match.

Note that this Unity change also means that Rider will break on unhandled exceptions that the user throws, which is a change in behaviour from previous versions. Ironically, this release of Rider suppresses unhandled exceptions in IL2CPP, which has always reported them to the debugger, to better match the old editor behaviour (and it also helps avoid situations with empty call stacks due to exceptions being thrown in library code that IL2CPP doesn't generate debug information for, and/or which doesn't have any managed code to decompile and show). This feature is enabled by default but can be disabled in the settings. We can re-evaluate the default setting for the next release, depending on what Unity decides should be their default unhandled exception strategy.


ArsenyChernyaev (approved):

krasnotsvetov (approved):

citizenmatt:
See also [RIDER-64944](https://youtrack.jetbrains.com/issue/RIDER-64944)

PR #2123, opened 2021-07-17
Fix "Show in Unity" for local packages

van800:
This action "Show in Unity" can be called directly on current active file via Shift-Shift or for some files we show a notification in the editor with link to "Show in Unity" action.

This PR fixes the action for local packages and also avoids showing a link in notification for "meta" files.

### DIFF
--- a/unity/EditorPlugin/AfterUnity56/Navigation/Initialization.cs
+++ b/unity/EditorPlugin/AfterUnity56/Navigation/Initialization.cs
@@ -2,6 +2,7 @@ using System;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using JetBrains.Collections.Viewable;
 using JetBrains.Rider.Model.Unity.BackendUnity;
 using JetBrains.Rider.Unity.Editor.Navigation;
 using JetBrains.Rider.Unity.Editor.Navigation.Window;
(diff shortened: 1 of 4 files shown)


krasnotsvetov (approved):

unity/EditorPlugin/AfterUnity56/Navigation/Initialization.cs
```diff
@@ -63,16 +65,17 @@ public static void OnModelInitializationHandler(UnityModelAndLifetime modelAndLi
         }
       });
 
-      modelValue.ShowFileInUnity.Advise(connectionLifetime, result =>
+      modelValue.ShowFileInUnity.AdviseNotNull(connectionLifetime, result =>
       {
-        if (result != null)
+        var matchedUnityPath = AssetDatabase.GetAllAssetPaths()
```
citizenmatt, line 70: This could be a very expensive call, for very large projects. We could get an array of tens of thousands of asset paths. We're also creating 2 `FileInfo` instances for each asset. Could we do this as a fallback if we can't get the path another way?

van800: That is definitely possible. I will restart the review again.

van800: https://github.com/JetBrains/resharper-unity/pull/2124

van800:
This was reverted in favor of https://github.com/JetBrains/resharper-unity/pull/2124

PR #2124, opened 2021-07-19
Fix "Show in Unity" for local packages

van800:
This action "Show in Unity" can be called directly on current active file via Shift-Shift or for some files we show a notification in the editor with link to "Show in Unity" action.

This PR fixes the action for local packages and also avoids showing a link in notification for "meta" files.
https://youtrack.jetbrains.com/issue/RIDER-65128


citizenmatt (approved):

van800:
We have found a behaviour of Path.GetFullPath, which brakes logic in this PR
image

PR #2134, opened 2021-08-02
Show in unity fix for `package@version` in path

van800:
see https://github.com/JetBrains/resharper-unity/pull/2124#issuecomment-890868130

Planned for backport to 212.


citizenmatt:
We have all the information to know if a file is part of a package. Would it help if we could get `package@version` to provide a project-relative path that would work with the asset DB without having to load everything?

van800:
We have to load it anyway to highlight in Unity Editor.

citizenmatt:
Sorry, I meant we could construct the `package@version` path so we could avoid calling `GetAllAssetPaths` and do a string comparison for each path.

van800:
Yes, I see what you mean now. I should do a measurement on a really big solution, to see, if we need such optimisation.

PR #2145, opened 2021-08-11
new Frontend meta-file-tracker

van800:
Need to mention that:
1. for ReSharper plugin we are dropping the MetaFileTracker, because VSTU has one already.
2. Undo would not work in cases when both Rider and Unity change csproj file. But we have decided it is not so important for now.


resharper/resharper-unity/src/Rider/UnityRefreshTracker.cs
```diff
@@ -215,7 +215,7 @@ private void FileSystemChange(FileSystemChange fileSystemChange)
 
         private void AdviseFileAddedOrDeleted(FileSystemChangeDelta delta)
         {
-            if (delta.NewPath.ExtensionNoDot == "cs")
+            if (delta.NewPath.ExtensionNoDot == "cs" || delta.NewPath.ExtensionNoDot == "asmdef")
```
krasnotsvetov, line 218: btw, what's about .shader, .cginc, .cg and other files which are added to csproj generation at unity side?

van800: This is done specifically for the case, when csproj didn't exist and should be created when asmdef or cs file is added. Not sure, if the same is really needed for extensions, which you have mentioned. Still focusing unity or calling Save in Rider would also trigger the same project generation. So it is a matter of convenience for a special case.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/explorer/MetaTracker.kt
```diff
@@ -0,0 +1,226 @@
...
+        }
+
+        if (actions.isEmpty()) return
+
+        val commandProcessor = CommandProcessor.getInstance()
+        var groupId = commandProcessor.currentCommandGroupId
+        if (groupId == null) {
+            groupId = MetaGroupId(nextGroupIdIndex++)
+            commandProcessor.currentCommandGroupId = groupId
+        }
```
citizenmatt, line 124: What does the group ID do?

citizenmatt (approved):
Looks ok to me. Added some comments/questions, no showstoppers, though. Might be worth adding a couple of comments or renaming methods.

I'm happy to go with this, as long as we're happy that meta files will still be correctly updated in Visual Studio with ReSharper doing refactorings, etc.

PR #2160, opened 2021-08-31
RIDER-67197 Add fix for instrumentation task classpath.

mfilippov:


citizenmatt (approved):
I don't really know what this is doing. We had `intellij.instrumentCode = false` in the old `.gradle` files. But either way, LGTM.

PR #2171, opened 2021-09-15
Refactor external asset file module

citizenmatt:
This PR is mostly a refactoring of the existing external files module for indexing meta and asset files, in preparation for also indexing `.asmdef` files.

* Moves the external files PsiModule from `resharper-unity/src/Yaml` to a new top level `resharper-unity/src/Core` folder, and updates namespaces. Since the module will no longer just be YAML files, but also JSON files, it makes sense to move it out of the `Yaml` folder/namespace.
    * The top level `Core` folder is to help tidy up the top level folder structure of the project. We've got language specific folders, such as `CSharp`, `Yaml`, `ShaderLab` and `HlslSupport`, etc. but we don't have anything for components that are not language specific. Instead we have folders such as `Feature`, `Application`, `Debugger`, etc. Future changes will move more of these folders into `Core`.
* Moves the `IProjectFile` creation for Rider out of `#if RIDER` blocks and into a separate component
* Removes the obsolete `EnableYamlParsing2` setting. Nothing was ever setting this value. If we want to disable parsing, we can use the checkbox in settings to disable indexing instead.
* Simplifies the checks for project settings analysis by moving the checks into `UnityProjectSettingsCache.IsAvaialble`
* Everything else is just files moves, renames and namespace updates

This PR is currently based on #2169, but will be retargeted to `net213` once that's merged.


resharper/resharper-unity/src/Yaml/Psi/DeferredCaches/UnityAssetsCacheUpdater.cs
```diff
@@ -26,14 +28,16 @@ public UnityCacheUpdater(Lifetime lifetime, ISolution solution, DeferredCacheCon
...
                     {
                         var psiSourceFiles = module.SourceFiles.ToList();
                         if (psiSourceFiles.Any())
                         {
                             locks.ExecuteWithWriteLock(() =>
                             {
                                 deferredCacheController.Invalidate<UnityAssetsCache>();
+
+                                // TODO: Why is this being run through ChangeManager?
                                 changeManager.ExecuteAfterChange(() =>
```
krasnotsvetov, line 41: Which way will be better for updating SimpleICaches for asset source files?

krasnotsvetov (approved):

resharper/resharper-unity/src/Yaml/Psi/DeferredCaches/UnityAssetsCacheUpdater.cs
citizenmatt: The update bit is fine, I don't get why it's done inside `ExecuteAfterChange`. I don't see how this could be called while executing another change.

PR #2174, opened 2021-09-28
Process files from all external packages

citizenmatt:
This PR will:

* Add all external packages to the external files module. Previously, we would process `Assets`, `ProjectSettings` and `Packages`. This would miss all `file:` based packages, as well as read only, git and built in packages. This means that the meta file cache now includes all meta files from all packages, and we'll build indexes of all assets, prefabs and scenes (See [RIDER-66223](https://youtrack.jetbrains.com/issue/RIDER-66223))
* Folders ending with `~` are not processed. These are not imported into the Unity asset database, and so are not treated as part of the project. They are usually used for `Samples~` and `Documentation~`.
* Add `.asmdef` files to the external files module. This means the asmdef name cache contains all assembly definition names in the project, not just from editable packages and `Assets`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ Since 2018.1, the version numbers and release cycle match Rider's versions and r
 
 - Process external packages for assets, `.meta` and `.asmdef` files ([RIDER-66223](https://youtrack.jetbrains.com/issue/RIDER-66223), [#2174](https://github.com/JetBrains/resharper-unity/pull/2174))
 - Update Unity API information to 2021.2.0b8 ([#2151](https://github.com/JetBrains/resharper-unity/pull/2151))
-- Rider: Several fixes and improvements to meta file tracking ([RIDER-48085](https://youtrack.jetbrains.com/issue/RIDER-48085), [RIDER-53643](https://youtrack.jetbrains.com/issue/RIDER-53643), [#2145](https://github.com/JetBrains/resharper-unity/pull/2145))
+- Rider: Several fixes and improvements to meta file tracking ([RIDER-48085](https://youtrack.jetbrains.com/issue/RIDER-48085), [RIDER-53643](https://youtrack.jetbrains.com/issue/RIDER-53643), [RIDER-67666](https://youtrack.jetbrains.com/issue/RIDER-67666), [#2145](https://github.com/JetBrains/resharper-unity/pull/2145))
 - ReSharper: Improve asset usage grouping ([#2169](https://github.com/JetBrains/resharper-unity/pull/2169))
 - ReSharper: Remove meta file tracking in favour of VSTU's implementation ([RIDER-48085](https://youtrack.jetbrains.com/issue/RIDER-48085), [#2145](https://github.com/JetBrains/resharper-unity/pull/2145))
 


resharper/resharper-unity/src/Core/Feature/Services/UsageStatistics/UnityExternalFilesFileSizeLogContributor.cs
```diff
@@ -12,33 +14,94 @@ public class UnityExternalFilesFileSizeLogContributor : IActivityLogContributorS
...
+            stats["rad"] = JArray.FromObject(GetStatistics(FileType.AsmDef, false));
+            stats["rkba"] = JArray.FromObject(GetStatistics(FileType.KnownBinary, false));
+            stats["rebna"] = JArray.FromObject(GetStatistics(FileType.ExcludedByName, false));
+
+            // There's no need to capture all meta file sizes. Just get a count and an average size
+            stats["mfc"] = myStatistics.Count(d => d.FileType == FileType.Meta && d.IsUserEditable);
+            stats["rmfc"] = myStatistics.Count(d => d.FileType == FileType.Meta && !d.IsUserEditable);
+
+            // "All meta file average"
+            stats["amfa"] = myStatistics.Average(d => d.FileType == FileType.Meta ? d.Length : 0.0f);
```
krasnotsvetov, line 62: Average is not correct, the count is not changed, but some elements contribute 0

citizenmatt: Oops. Yes, good catch. I'll get it fixed. Are you happy with the statistics that are there?

krasnotsvetov: Yes

resharper/resharper-unity/src/Core/Psi/Modules/UnityExternalFilesModuleProcessor.cs
```diff
@@ -179,45 +145,98 @@ private void CollectExternalFilesForDirectory(ExternalFiles externalFiles, Virtu
             // Mac, on the same project!
             var entries = directory.GetDirectoryEntries("*", PathSearchFlags.RecurseIntoSubdirectories
                                                              | PathSearchFlags.ExcludeDirectories);
+
             foreach (var entry in entries)
-                externalFiles.AddFile(entry);
+            {
+                // Ignore anything under a folder that ends with a tilde - Unity does not import these folders into the
+                // asset database, so they're not part of this project
+                if (entry.IsFile && !entry.RelativePath.FullPath.Contains("~/") &&
```
krasnotsvetov, line 153: What do you think about using GetDirectoryEntries without recursive mode and manually enter directories which name does not contain `~`? In that case we will scan shorter lines + we will not scan child files inside directory with `~` at all

resharper/resharper-unity/src/Core/Psi/Modules/UnityExternalFilesModuleProcessor.cs
```diff
@@ -25,148 +30,109 @@ public class UnityExternalFilesModuleProcessor : IChangeProvider, IUnityReferenc
...
-                        return;
-                    }
-                }
-            }
+                CollectExternalFilesForDirectory(externalFiles, path, true);
         }
 
-        private void CollectExternalFilesForDirectory(ExternalFiles externalFiles, VirtualFileSystemPath directory)
+        private void CollectExternalFilesForDirectory(ExternalFiles externalFiles, VirtualFileSystemPath directory,
+                                                      bool isUserEditable)
```
krasnotsvetov, line 119: Let's add assert that directory is absolute path

resharper/resharper-unity/src/Core/Psi/Modules/UnityExternalFilesModuleProcessor.cs
citizenmatt: Hmm. The majority of files will be without the `~`, but it's probably worth unrolling it.

krasnotsvetov (changes requested):
What will happen with .asmdef file if it was handled by module processor while packages was not included to project generation and after that user has enabled project generation for that package? I do not see code which handles that case

resharper/resharper-unity/src/Core/Psi/Modules/UnityExternalFilesModuleProcessor.cs
krasnotsvetov: Yes, the majority of files will be without `~` and for each file we will scan the whole full path string for `~` instead of short name of current file or directory, e.g let's consider several paths:
`PackageName/Assets/Folder1/Foo1/Foo2/Name1/`
...
`PackageName/Assets/Folder1/Foo1/Foo2/Name100/`

You will scan 100 full pathes for `~` instead of scanning `PackageName`, `Assets`, `Folder1`, `Foo1`, `Foo2` and names since `Name1` up to `Name100`. 105 short names instead of 100 full pathes. But possibly it will be fast.

krasnotsvetov: But I am not sure that unrolling GetDirectoryEntries will be same to GetDirectoryEntries with recursive flag, we could lost all benefits from unrolling if it is true

PR #2201, opened 2021-11-01
fix RIDER-69804

van800:


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/toolWindow/log/UnityLogPanelModel.kt
```diff
@@ -191,5 +192,17 @@ class UnityLogPanelModel(lifetime: Lifetime, val project: Project, toolWindow: T
...
         project.solution.frontendBackendModel.consoleLogging.lastPlayTime.advise(lifetime){ queueUpdate() }
+        project.messageBus.connect(toolWindow.contentManager).subscribe(
+            ToolWindowManagerListener.TOPIC,
+            createToolwindowManagerListener())
+    }
+
+    private fun createToolwindowManagerListener(): ToolWindowManagerListener {
+        return object : ToolWindowManagerListener {
+            override fun toolWindowShown(toolWindow: ToolWindow) {
+                mergingUpdateQueueAction.run()
```
ForNeVeR, line 203: Please check whether the `toolWindow` in question is the one you want (like `toolWindow === this@UnityLogPanelModel.toolWindow`, I think?)

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/toolWindow/log/UnityLogPanelModel.kt
```diff
@@ -191,5 +192,17 @@ class UnityLogPanelModel(lifetime: Lifetime, val project: Project, toolWindow: T
...
+                super.toolWindowShown(toolWindow)
```
ForNeVeR, line 204: _(Minor)_ `super` call is not technically required here, super method is an empty default interface method.

ForNeVeR (changes requested):

ForNeVeR (approved):

PR #2205, opened 2021-11-03
fix RIDER-70098 - MetaTracker for multiple opened solutions

van800:
Correct MetaTracker for the case with multiple opened Unity/non-Unity solutions

### DIFF
--- a/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/explorer/MetaTracker.kt
+++ b/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/explorer/MetaTracker.kt
@@ -5,18 +5,21 @@ package com.jetbrains.rider.plugins.unity.explorer
 import com.intellij.openapi.Disposable
 import com.intellij.openapi.command.CommandProcessor
 import com.intellij.openapi.command.undo.UndoManager
+import com.intellij.openapi.components.serviceIfCreated
 import com.intellij.openapi.project.Project
+import com.intellij.openapi.project.ProjectManager
 import com.intellij.openapi.vfs.LocalFileSystem
 import com.intellij.openapi.vfs.VfsUtil
 import com.intellij.openapi.vfs.VirtualFile
-import com.intellij.openapi.vfs.VirtualFileManager
 import com.intellij.openapi.vfs.newvfs.BulkFileListener
 import com.intellij.openapi.vfs.newvfs.events.*
 import com.intellij.util.PathUtil
 import com.intellij.util.application
+import com.intellij.workspaceModel.ide.WorkspaceModel
 import com.jetbrains.rd.platform.util.getLogger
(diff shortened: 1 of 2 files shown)


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/explorer/MetaTracker.kt
```diff
@@ -30,109 +32,105 @@ import kotlin.io.path.ExperimentalPathApi
...
-                    actions.executeUnderWriteLock()
-                }, actions.getCommandName(), groupId)
+            application.invokeLater {
+                commandProcessor.allowMergeGlobalCommands {
+                    commandProcessor.executeCommand(project, {
+                        actions.executeUnderWriteLock()
+                    }, actions.getCommandName(), groupId)
+                }
             }
+            return // action was handled, so we should not try to process it again with other projects
```
citizenmatt, line 129: There's an edge case here, if multiple Unity projects are open and you somehow get events from both projects. This is a pretty small edge case though. I'm finding it hard to think of a good scenario.

Perhaps a `file:` package added to both Project A and Project B, and a bulk update is done for files in project B. Project A would process the shared file, then return. The changes for other files in project B would be missed. I think we'd be fine with this, unless you can think of a more likely scenario?

citizenmatt (commented):
LGTM mostly. Check out the comment, see what you think.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/explorer/MetaTracker.kt
van800: I suppose this scenario would be fine, since we process each event separately for every project opened.

van800: The service is application level, it gets all the changes and trues to find appropriate unity project (solution) for each change. Once the event is processed with any of the projects, it stops trying the others.

citizenmatt: > I suppose this scenario would be fine, since we process each event separately for every project opened.

It looks like it's processing a list of events by looping over each project, and if that project processes at least one event, then it stops looping. This relies on the list of events all belonging to one project, doesn't it?

van800: You are right, it looks like a bug. investigating...

van800: @citizenmatt Please take a look at this new pull request https://github.com/JetBrains/resharper-unity/pull/2208/files

PR #2207, opened 2021-11-05
Fix issues with iOS USB device enumeration app

citizenmatt:
This PR fixes some issues with the helper app that's invoked when the Attach to Unity Process dialog starts. This app loads a native library provided by Unity and polls for Apple devices attached via USB. If the iOS Build Support module isn't installed, the helper app isn't invoked.

- [x] Stop throwing exceptions on a background thread. Fixes [RIDER-69809](https://youtrack.jetbrains.com/issue/RIDER-69809). Some versions of macOS will silently ignore this, others will show a "dotnet has unexpectedly quit" dialog
- [x] Update paths to native library for Unity 2021.2. The files have moved into `x64` and `arm64` directories. The app will now try to load the library from the new location and fallback to the location from previous versions
- [x] Load the correct native library when running on M1 Macs. If Rider is running under Rosetta, the helper app will load the `x64` library. If running natively on M1, it will load `arm64`. If running natively on M1 and there is no native install of Unity, it will gracefully not load the library and USB debugging will be unavailable. The [troubleshooting guide](https://github.com/JetBrains/resharper-unity/wiki/Troubleshooting-debugging-Unity-players) linked from the dialog has been updated to include this scenario. (Loading the wrong library is what caused the exception shown in [RIDER-69809](https://youtrack.jetbrains.com/issue/RIDER-69809))
- [x] Support USB debugging on Linux, loading the native library that's available on Linux (and correctly handling the path change for Unity 2021.2)

Note that the change in location for Unity 2021.2 means that older versions of Rider will be unable to load the library with newer versions of Unity, and USB debugging will not be available. Depending on OS, there might also be a crash dialog shown.

### DIFF
--- a/debugger/ios-list-usb-devices/src/Program.cs
+++ b/debugger/ios-list-usb-devices/src/Program.cs
@@ -1,14 +1,13 @@
 ﻿using System;
 using System.Net.Sockets;
 using System.Threading;
+using System.Threading.Tasks;
 
 namespace JetBrains.Rider.Plugins.Unity.iOS.ListUsbDevices
 {
     internal static class Program
     {
-        private static bool ourFinished;
-
-        private static int Main(string[] args)
+        private static async Task<int> Main(string[] args)
         {
             if (args.Length != 2)
             {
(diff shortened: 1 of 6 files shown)


debugger/ios-list-usb-devices/src/Program.cs
```diff
@@ -19,21 +18,34 @@ private static int Main(string[] args)
...
             InitialiseWinSock();
 
-            var thread = new Thread(ThreadFunc);
-            thread.Start(args);
+            var iosSupportPath = args[0];
+            var pollingInterval = TimeSpan.FromMilliseconds(int.Parse(args[1]));
+
+            var pollingCancellationSource = new CancellationTokenSource();
+            var readLineTask = CreateAwaitStopCommandTask();
+            var innerPollingTask = CreatePollForDevicesTask(iosSupportPath, pollingInterval,
```
van800, line 26: var innerPollingTask = await CreatePollForDevicesTask(iosSupportPath, pollingInterval,
                pollingCancellationSource.Token);

and no need to await below.

citizenmatt: I need to await both the read from the command line and the poll at the same time. If the read comes in, it needs to cancel the poll. Without the `ContinueWith`, the final await will throw an unhandled `OperationCancelledException`, and any exception thrown by the poll is unobserved and doesn't get reported

van800: got it, thanks

Issue #2199, opened 2021-10-27
Usages of JetBrains.Annotations and API verifier


ulex:
Hey
currently we are working on api verifier, which should help plugin writers to avoid updating plugin on every new release of ReSharer/Rider SDK, when they are binary compatible.
I've noticed, that Unity plugin already use forward-compability version specification format, which is well suits for api verifier testing purposes, thanks.

The question is about usages of JetBrains.Annotations.
Usually, there are no problem of using annotations, since they are marked by `Conditional` attribute, and by default their usages are not presented in the resulting dlls.
But I've noticed, that the plugin directly uses types from JetBrains.Annotations package for semantically different purposes, like here:
https://github.com/JetBrains/resharper-unity/blob/net213/resharper/resharper-unity/src/CSharp/Daemon/UsageChecking/UsageInspectionsSuppressor.cs#L41
And like this
```
private static readonly IClrTypeName ourPublicAPIAttribute = (IClrTypeName) new ClrTypeName(typeof (PublicAPIAttribute).FullName, (IList<int>) null);
```

JetBrains.Annotation is the only package (I hope) where version isn't fixed, like in our other sdk-libraries. And for that reason every, even smallest update of JetBrains.Annotations in SDK will break API compability for resharper-unity plugin.

So I propose either:
1. Copy annotations to plugin
2. Replace existing usages to constants. Make copy of `ImplicitUseKindFlags`

What do you think?

citizenmatt:
We can easily change the usage of `typeof(PublicAPIAttribute).FullName`, but we can't easily change `ImplicitUseKindFlags`. That comes from `IUsageInspectionsSuppressor.SuppressUsageInspectionsOnElement`, which is an API from the platform, defined in `JetBrains.ReSharper.Daemon.Engine.dll`. We need to implement that, so we'll need to match the type signature, which AIUI means we need to reference `JetBrains.Annotations.dll`. I'm not sure where the reference to the annotations package comes from - I can't see it explicitly, so I guess it's a transitive dependency from the SDK?

ulex:
Wow, indeed. Sorry, didn't noticed that. I will try to modify this and related APIs and will reopen this issue then. Thanks!

Yes, the JetBrains.Annotations comes from SDK, for example psi.features uses it. 
https://www.nuget.org/packages/JetBrains.Psi.Features.src/213.0.20211012.94415-eap04

PR #2209, opened 2021-11-09
Drop usages of TestRunner.Abstractions

ulex:
Related to  #2199

`TestRunner.Abstractions` has copies of NotNull and similar extension methods, but have a fixed version and any usage of its members will raise an error in runtime when the product is updated.


citizenmatt (approved):
LGTM!

PR #2223, opened 2021-11-26
Notification on Rider package update availability

van800:
@citizenmatt what do you think?

The idea is to have both hard-coded expected version of the Rider package (**3.0.7**) and if protocol is available, we call the code on the EditorPlugin side to get more accurate newest compatible expected version.


resharper/resharper-unity/src/Rider/RiderPackageUpdateAvailabilityChecker.cs
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        private void ShowNotificationIfNeeded(Lifetime lifetime)
+        {
+            if (!myBoundSettingsStore.GetValue((UnitySettings s) => s.AllowRiderUpdateNotifications))
+                return;
+            
+            myPackageManager.IsInitialUpdateFinished.WhenTrue(lifetime, lt =>
+            {
+                var version = myUnityVersion.ActualVersionForSolution.Value;
```
citizenmatt, line 80: Might be worth combining `ActualVersionForSolution` with `IsInitialUpdateFinished`. When both are true/valid, then run the check

resharper/resharper-unity/src/Rider/RiderPackageUpdateAvailabilityChecker.cs
```diff
@@ -0,0 +1,127 @@
...
+                    var notification = new NotificationModel($"JetBrains Rider package in Unity is missing.", "Make sure JetBrains Rider package is installed in Unity Package Manager.", true, RdNotificationEntryType.WARN, new List<NotificationHyperlink>());
+                    myShellLocks.ExecuteOrQueueEx(lt, "RiderPackageUpdateAvailabilityChecker.ShowNotificationIfNeeded", () => myNotifications.Notification(notification));
+                }
+                // todo: read available newest compatible Rider package version from somewhere, like protocol or json file
+                else if (package.Source == PackageSource.Registry && new Version(package.PackageDetails.Version) < new Version(3,0,7))
+                {
+                    myNotificationShown.Add(mySolution.SolutionFilePath);
+                    myLogger.Info($"{packageId} {package.PackageDetails.Version} is older then expected.");
+                    var neverShowLink = new NotificationHyperlink("Never show for this solution", myHyperlinkId);
+                    var notificationId = 123409;
```
citizenmatt, line 106: Where does this magic number come from?

resharper/resharper-unity/src/Rider/RiderPackageUpdateAvailabilityChecker.cs
```diff
@@ -0,0 +1,127 @@
...
+                    var notification = new NotificationModel("Update available - JetBrains Rider package.", 
+                        "Check for JetBrains Rider package updates in Unity Package Manager.", 
+                        true, RdNotificationEntryType.INFO, new List<NotificationHyperlink> {neverShowLink}, notificationId);
+                    var notificationLifetime = lt.CreateNested();
+                    myNotifications.ExecuteHyperlink.Advise(notificationLifetime.Lifetime, id =>
+                    {
+                        if (id == myHyperlinkId)
+                        {
+                            mySettingsStore.BindToContextTransient(ContextRange.ManuallyRestrictWritesToOneContext(mySolution.ToDataContext()))
+                                .SetValue((UnitySettings key) => key.AllowRiderUpdateNotifications, false);
```
citizenmatt, line 116: This writes to `.sln.DotSettings`, doesn't it? Might be better to use the persistent solution properties, which are stored in the ReSharper cache folder, rather than the user's settings.

citizenmatt:
Nice! I like the idea of checking with the registry - that helps with the scenario of a newer Unity and older Rider.

This will show the notification in Rider. Are we interested in showing it in Unity at all? I'm not sure if they have a notifications API, but we can add something to the log? Or do you think it's best to just keep it in Rider?

resharper/resharper-unity/src/Rider/RiderPackageUpdateAvailabilityChecker.cs
van800: It is a random number. Any number except conflicting with other notifications would work.
I asked in https://jetbrains.slack.com/archives/CH9S2GGRG/p1637923373479800, it seems it is good enough. But I will evaluate other suggestions.

van800:
> Nice! I like the idea of checking with the registry - that helps with the scenario of a newer Unity and older Rider.
> 
> This will show the notification in Rider. Are we interested in showing it in Unity at all? I'm not sure if they have a notifications API, but we can add something to the log? Or do you think it's best to just keep it in Rider?

I want to keep it in Rider only.

resharper/resharper-unity/src/Rider/RiderPackageUpdateAvailabilityChecker.cs
van800: It goes to .sln.dotSettings.user

krasnotsvetov: Please store that in solution caches, do not modify anything in project on openning

van800: If I start using persistent solution properties, how user would enable it back if occasionally turn off with the link in the popup? In the current implementation, I have a setting with a checkbox in the Preferences.

PR #2229, opened 2021-12-13
Improvment for RIDER-71633, add logging for RIDER-71921

krasnotsvetov:
RIDER-71633 will be fixed in 22.1, I guess we should add files and folders to ignore or reuse existing one for code analysis


resharper/resharper-unity/src/Feature/Caches/DeferredCacheController.cs
```diff
@@ -167,6 +167,7 @@ public bool IsProcessingFiles()
                 {
                     // assertion is incorrect, because file could be removed while we have read lock.
                     // Assertion.Assert(!myDeferredHelperCache.FilesToProcess.Contains(psiSourceFile), "!myDeferredHelperCache.FilesToProcess.Contains(psiSourceFile), removed");
+                    myLogger.Verbose("Remove from processing removed file: {0}", psiSourceFile.GetPersistentIdForLogging());
```
citizenmatt, line 170: Can the message be updated to better reflect that the file has been (unexpectedly) deleted, please?

citizenmatt (approved):
Minor change for log message, but LGTM :)

PR #2231, opened 2021-12-15
Generate run configuration for batchmode tests

van800:
fix https://youtrack.jetbrains.com/issue/RIDER-70675

### DIFF
--- a/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/util/UnityUtils.kt
+++ b/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/util/UnityUtils.kt
@@ -54,7 +54,7 @@ fun MutableList<String>.withRunTests(project: Project) : MutableList<String> {
 }
 
 fun MutableList<String>.withTestResults(project: Project) : MutableList<String> {
-    this.addAll(listOf("-testResults", project.solutionDirectory.resolve("results.xml").canonicalPath))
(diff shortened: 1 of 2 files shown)


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/util/UnityUtils.kt
```diff
@@ -43,6 +43,21 @@ fun MutableList<String>.withProjectPath(project: Project) : MutableList<String>
...
+    return this
+}
+
+fun MutableList<String>.withRunTests(project: Project) : MutableList<String> {
+    this.add("-runTests")
+    return this
+}
+
+fun MutableList<String>.withTestResults(project: Project) : MutableList<String> {
+    this.addAll(listOf("-testResults", project.solutionDirectory.resolve("results.xml").canonicalPath))
```
krasnotsvetov, line 57: I am not sure that results.xml in repo root is a good place, what's about creating seperated folder. e.g UnitTestsResults? Or put results in library

krasnotsvetov (approved):

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/util/UnityUtils.kt
van800: I have changed it to point to `Temp` folder in the solution. That Temp folder is created anyway by Unity.

PR #2235, opened 2022-01-05
improve heuristics for checking that project is Unity

van800:
Be more strict in checking that a project is Unity one.
Initialize less features for a folder which looks like Unity.
Should help with https://youtrack.jetbrains.com/issue/RIDER-70936


resharper/resharper-unity/src/Unity/Core/ProjectModel/UnitySolutionTracker.cs
```diff
@@ -41,7 +41,7 @@ public UnitySolutionTracker(ISolution solution, IFileSystemTracker fileSystemTra
         private void SetValues()
         {
             IsUnityProjectFolder.SetValue(HasUnityFileStructure(mySolution.SolutionDirectory));
-            IsUnityProject.SetValue(IsUnityProjectFolder.Value && mySolution.IsValid() && mySolution.SolutionFilePath.ExistsFile);
+            IsUnityProject.SetValue(IsUnityProjectFolder.Value && mySolution.IsValid() && mySolution.SolutionFilePath.ExistsFile && HasLibraryFolder(mySolution.SolutionDirectory));
```
citizenmatt, line 44: Should `HasLibraryFolder` be part of `HasUnityFileStructure`?

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/UnityProjectDiscoverer.kt
```diff
@@ -58,6 +58,12 @@ class UnityProjectDiscoverer(private val project: Project) : LifetimedService()
         }
     }
 
+    // When Unity has generated sln, Library folder was also created. Lets' be more strict and check it.
+    private fun hasLibraryFolder(project: Project): Boolean {
```
citizenmatt, line 62: Same as on backend, should this be part of `hasUnityFileStructure`?

resharper/resharper-unity/src/Unity/Core/ProjectModel/UnitySolutionTracker.cs
van800: No, because Library is usually not pushed to vcs. After checking out project from vcs, we need to show notification/action to start unity. After solution is generated, Library folder also already exist.

Issue #2164, opened 2021-09-03
Consolidate UnityReferencesTracker and UnitySolutionTracker


citizenmatt:
Both classes can be used to track if a solution is a Unity project or not. `UnitySolutionTracker` is based on file system. `UnityReferencesTracker` is based on references. It's not clear when to use one instead of the other.

Also, it is very easy to create circular references with `UnityReferencesTracker`. It takes in implementations of `IUnityReferenceChangeHandler` for push notifications when references change (or are first read), but also exposes `HasUnityReference` for pull notifications (technical push or pull, as it's a `ViewableProperty`, but it's most frequently used to check `HasUnityReference.Value`.

Consolidate these classes, and reduce the amount of classes passed as dependencies. `IUnityReferenceChangeHandler` should either be controlled by another component, or the implementations should receive `UnityReferencesTracker` and subscribe to viewable properties/collections.

PR #2243, opened 2022-01-14
refactor UnityReferencesTracker and UnitySolutionTracker

van800:
fix https://youtrack.jetbrains.com/issue/RIDER-69292
fix https://github.com/JetBrains/resharper-unity/issues/2164

### DIFF
--- a/resharper/resharper-unity/src/Unity.Shaders/HlslSupport/Integration/Cpp/UnityShaderModuleProvider.cs
+++ b/resharper/resharper-unity/src/Unity.Shaders/HlslSupport/Integration/Cpp/UnityShaderModuleProvider.cs
@@ -118,17 +118,17 @@ namespace JetBrains.ReSharper.Plugins.Unity.HlslSupport.Integration.Cpp
     [SolutionComponent]
     public class UnityShaderPsiModuleProviderFilter : IProjectPsiModuleProviderFilter
     {
-        private readonly UnitySolutionTracker mySolutionTracker;
+        private readonly UnityReferencesTracker myTracker;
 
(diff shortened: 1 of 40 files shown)


resharper/resharper-unity/src/Unity.Shaders/HlslSupport/Integration/Cpp/UnityShaderModuleProvider.cs
```diff
@@ -118,17 +118,10 @@ public string GetPersistentID()
...
-        public UnityShaderPsiModuleProviderFilter(UnitySolutionTracker solutionTracker)
-        {
-            mySolutionTracker = solutionTracker;
-        }
-
         public Tuple<IProjectPsiModuleHandler, IPsiModuleDecorator> OverrideHandler(Lifetime lifetime, IProject project,
             IProjectPsiModuleHandler handler)
         {
-            if ( handler.PrimaryModule != null && UnityReferencesTracker.ReferencesUnity(project))
+            if ( handler.PrimaryModule != null && project.GetComponent<UnityReferencesTracker>().IsUnityProject(project))
```
citizenmatt, line 124: Can we inject this instead of calling `GetComponent`?

citizenmatt (changes requested):
LGTM

resharper/resharper-unity/src/Unity.Shaders/HlslSupport/Integration/Cpp/UnityShaderModuleProvider.cs
van800: Tried that - caused cyclic dependencies, if I remember correct.

PR #2251, opened 2022-01-24
fix for possible location of local unity documentation

van800:
relates to https://github.com/JetBrains/resharper-unity/issues/306

### DIFF
--- a/resharper/resharper-unity/src/Unity/Core/Application/UI/Help/ShowUnityHelp.cs
+++ b/resharper/resharper-unity/src/Unity/Core/Application/UI/Help/ShowUnityHelp.cs
@@ -89,8 +89,10 @@ namespace JetBrains.ReSharper.Plugins.Unity.Core.Application.UI.Help
             var root = contentsPath.Combine("Documentation");
 
             // I see /home/ivan-shakhov/Unity/Hub/Editor/2021.2.4f1/Editor/Data/Documentation/Documentation/en path on my machine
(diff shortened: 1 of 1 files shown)


resharper/resharper-unity/src/Unity/Core/Application/UI/Help/ShowUnityHelp.cs
```diff
@@ -87,6 +87,11 @@ private FileSystemPath GetDocumentationRoot()
             var appPath = mySolutionsManager.Solution?.GetComponent<UnityVersion>().GetActualAppPathForSolution();
             var contentsPath = UnityInstallationFinder.GetApplicationContentsPath(appPath);
             var root = contentsPath.Combine("Documentation");
+
+            // I see /home/ivan-shakhov/Unity/Hub/Editor/2021.2.4f1/Editor/Data/Documentation/Documentation/en path on my machine
+            if (root.Combine("Documentation").ExistsDirectory)
+                root = root.Combine("Documentation");
```
citizenmatt, line 93: Very minor nitpick - let's pull `root.Combine()` out into a variable, so we don't have to parse and create the path twice.

citizenmatt (approved):
LGTM. Is this just a Linux thing? I don't see the extra path on macOS

van800:
> LGTM. Is this just a Linux thing? I don't see the extra path on macOS

Seems like Linux specific

PR #2253, opened 2022-01-25
fix a case when Attach to UnityEditor sun-configuration doesn't work

van800:
because of old serialized value of run configuration param


citizenmatt (approved):
LGTM!

PR #2266, opened 2022-02-03
fix RIDER-69350 Add a warning: MenuItem should be applied only to static methods

van800:
not sure, should I add "Why Re# is suggesting this" wiki page or it is obvious?


resharper/resharper-unity/test/src/Unity.Tests/Unity/CSharp/Daemon/Stages/Analysis/NonStaticMenuItemProblemAnalyzer.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace JetBrains.ReSharper.Plugins.Tests.Unity.CSharp.Daemon.Stages.Analysis
+{
+    [TestUnity]
+    public class NonStaticMenuItemProblemAnalyzerTest
+        : CSharpHighlightingTestBase<IncorrectSignatureWarning>
+    {
+        protected override string RelativeTestDataPath => @"CSharp\Daemon\Stages\Analysis";
+
+        [Test] public void TestInvalidStaticModifierWarning() { DoNamedTest2(); }
```
citizenmatt, line 12: Might be easier to just add this to `AttributedMethodSignatureAnalyzerTests`

citizenmatt (changes requested):
Looks like the `.gold` file is out of date, but other than that, LGTM

PR #2282, opened 2022-02-22
fix a case of navigation from LogView

van800:
In Unity code like
```
[InitializeOnLoad]
public class NewBehaviourScript 
{
static NewBehaviourScript()
{
    throw new Exception("");
}
}
```
produces log message like:
message = "System.Exception
  at NewBehaviourScript..cctor () [0x00001] in /home/ivan-shakhov/Work/com.unity.ide.rider/Assets/NewBehaviourScript.cs:13 "
stacktrace = "UnityEditor.EditorAssemblies:ProcessInitializeOnLoadAttributes (System.Type[]) (at /home/bokken/buildslave/unity/build/Editor/Mono/EditorAssemblies.cs:138)"

We need to use the message to navigate to the error.


van800:
Merged

citizenmatt:
Cherry picked as b5112bc13952187c439b1e80741e77ccb163c3b4

PR #2286, opened 2022-02-28
fix showing notification for the case when folder is opened instead o sln

van800:
image
image

### DIFF
--- a/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/OpenUnityProjectAsFolderNotification.kt
+++ b/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/OpenUnityProjectAsFolderNotification.kt
@@ -58,7 +58,6 @@ class OpenUnityProjectAsFolderNotification(project: Project) : ProtocolSubscribe
                         && !project.solution.frontendBackendModel.unityEditorConnected.valueOrDefault(false)
                         && !hasUnloadedProjects(project)
                     ) {
-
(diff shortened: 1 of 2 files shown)


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/notifications/OpenUnityProjectAsFolderNotification.kt
```diff
@@ -70,22 +70,28 @@ class OpenUnityProjectAsFolderNotification(project: Project) : ProtocolSubscribe
                 }
             }
             else if (solutionDescription is RdVirtualSolution) { // opened as folder
-                val adviceText = " Please <a href=\"close\">close</a> and reopen through the Unity editor, or by opening a .sln file."
+                val adviceText = "Please <a href=\"close\">close</a> and reopen through the Unity editor, or by opening a .sln file."
+                val mainText =
+                    if (solutionDescription.projectFilePaths.isEmpty())
+                        "This looks like a Unity project. C# and Unity specific features are not available when the project is opened as a folder."
+                    else
+                        "This looks like a Unity project. C# and Unity specific features are not available when only a single project is opened."
```
citizenmatt, line 78: How about we pull out "This looks like a Unity project" and make that the notification title for this scenario? It makes more sense than "Advanced Unity integration is unavailable"

citizenmatt (changes requested):

citizenmatt (approved):
LGTM!

PR #2325, opened 2022-06-14
fix RIDER-78230 Unity Play, Pause, Step buttons state is not saved on…

van800:
… IDE restart


citizenmatt:
Cherry picked as 55139e829e819b652e98f01590cfef17916e5a6b

PR #2326, opened 2022-06-16
Allow customizing "Start Unity" action by "Start Unity" run-configuration

van800:
Fix for:
https://youtrack.jetbrains.com/issue/RIDER-78619/Ability-to-set-custom-environment-variables-when-starting-Unity-from-Rider
https://youtrack.jetbrains.com/issue/RIDER-77371/Support-custom-Unity-command-line-arguments-in-Start-Unity-Rider-action

We are not removing the action only because it is not possible to assign shortcut for just run-configration.
https://youtrack.jetbrains.com/issue/IDEA-69968/Add-option-to-assign-keyboard-shortcut-to-RunDebug-configurations

### DIFF
--- a/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/actions/StartUnityAction.kt
+++ b/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/actions/StartUnityAction.kt
@@ -3,10 +3,9 @@ package com.jetbrains.rider.plugins.unity.actions
 import com.intellij.openapi.actionSystem.AnActionEvent
 import com.intellij.openapi.project.DumbAwareAction
 import com.intellij.openapi.project.Project
-import com.jetbrains.rider.plugins.unity.model.frontendBackend.frontendBackendModel
 import com.jetbrains.rider.plugins.unity.isConnectedToEditor
+import com.jetbrains.rider.plugins.unity.model.frontendBackend.frontendBackendModel
 import com.jetbrains.rider.plugins.unity.util.getUnityArgs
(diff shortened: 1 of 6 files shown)


rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/actions/StartUnityAction.kt
```diff
@@ -21,29 +29,48 @@ open class StartUnityAction : DumbAwareAction() {
...
+                    .withParentEnvironmentType(if (exeConfiguration.parameters.isPassParentEnvs) {
+                        GeneralCommandLine.ParentEnvironmentType.CONSOLE
+                    } else {
+                        GeneralCommandLine.ParentEnvironmentType.NONE
+                    })
+                    .withExePath(exeConfiguration.parameters.exePath)
+                    .withWorkDirectory(exeConfiguration.parameters.workingDirectory)
+                    .withRawParameters(exeConfiguration.parameters.programParameters)
+
+                return runCommandLine.toProcessBuilder().start()
```
ForNeVeR, line 62: Is there a reason you start the process manually here instead of using something like `ExecutionUtil.runConfiguration`?

That way, some details like before-launch-actions would be taken into account.

ForNeVeR (changes requested):

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/actions/StartUnityAction.kt
van800: https://jetbrains.slack.com/archives/C53QFRZ7G/p1655808937561529

PR #2337, opened 2022-07-13
fix DEXP-608179 One of the handlers for the BeforeAcquiringReadLock h…

van800:
…as failed. The execution context must be guarded from reentrancy in order to execute this action.


van800:
close in favor of https://jetbrains.team/p/net/reviews/5892/files

PR #2338, opened 2022-07-18
Pre commit hook, which was checking unsaved scenes, is not checking also other GameObjects states, like ScriptableObjects

van800:
Pre commit hook, which was checking unsaved scenes, is not checking also other GameObjects states, like ScriptableObjects.

### DIFF
--- a/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/ui/vcs/UnsavedSceneCheckinHandlerFactory.kt
+++ b/rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/ui/vcs/UnsavedSceneCheckinHandlerFactory.kt
@@ -32,14 +32,14 @@ private class UnresolvedMergeCheckHandler(
 ) : CheckinHandler() {
 
     private val project = panel.project
-    private val settings = UnsavedSceneCheckinState.getService(project)
+    private val settings = UnsavedCheckinState.getService(project)
     override fun getBeforeCheckinConfigurationPanel(): RefreshableOnComponent? {
         if (!project.isUnityProject())
             return null
 
         return BooleanCommitOption(
-            panel, UnityUIBundle.message("commitOption.check.unsaved.unity.scenes"), false,
-            settings::checkUnsavedScenes
+            panel, UnityUIBundle.message("commitOption.check.unsaved.unity.state"), false,
+            settings::checkUnsavedState
         )
     }
 
@@ -47,10 +47,10 @@ private class UnresolvedMergeCheckHandler(
         executor: CommitExecutor?,
         additionalDataConsumer: PairConsumer<Any, Any>
     ): ReturnResult {
-        if (settings.checkUnsavedScenes && project.isUnityProject()) {
+        if (settings.checkUnsavedState && project.isUnityProject()) {
             var providerResult = false
             try {
-                providerResult = project.solution.frontendBackendModel.hasUnsavedScenes
+                providerResult = project.solution.frontendBackendModel.hasUnsavedState
                     .sync(Unit, RpcTimeouts(200L, 200L))
             }
             catch (t: CancellationException){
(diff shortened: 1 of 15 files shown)


unity/EditorPlugin/AfterUnity56/UnitTesting/Initialization.cs
```diff
@@ -4,55 +4,115 @@
...
             {
-                if (SceneManager.GetSceneAt(i).isDirty)
-                    return true;
+                var methodInfo = typeof(EditorUtility).GetMethod("IsDirty",
+                    System.Reflection.BindingFlags.Static 
+                    | System.Reflection.BindingFlags.Public 
+                    | System.Reflection.BindingFlags.NonPublic, null, new[] { typeof(int)}, null);
+                if (methodInfo == null)
+                {
+                    ourLogger.Error("IsDirty method not found of type='{0}'", typeof(EditorUtility));
```
citizenmatt, line 103: Will this show in the Unity Console? Do we need it to be an error? We'll be checking this by default, so for (very) old Unity versions, we'd be showing an error that we can't do anything about.

van800: It is an error in the log file, not in the Unity Console.

unity/EditorPlugin/AfterUnity56/UnitTesting/Initialization.cs
```diff
@@ -4,55 +4,115 @@
...
+    public class SpawnManagerScriptableObject : ScriptableObject
     {
-        modelAndLifetime.Model.HasUnsavedScenes.Set(rdVoid => 
+        public string prefabName;
+    
+        public int numberOfPrefabsToCreate;
+        public Vector3[] spawnPoints;
+    }
+                 */
+                var hasDirtyUserAssets = Resources.FindObjectsOfTypeAll<Object>()
```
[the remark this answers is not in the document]
van800: IDK exactly, but only Resources.FindObjectsOfTypeAll works for my use-case.

rider/src/main/kotlin/com/jetbrains/rider/plugins/unity/ui/vcs/UnsavedSceneCheckinHandlerFactory.kt
```diff
@@ -32,25 +32,25 @@ private class UnresolvedMergeCheckHandler(
...
         executor: CommitExecutor?,
         additionalDataConsumer: PairConsumer<Any, Any>
     ): ReturnResult {
-        if (settings.checkUnsavedScenes && project.isUnityProject()) {
+        if (settings.checkUnsavedState && project.isUnityProject()) {
             var providerResult = false
             try {
-                providerResult = project.solution.frontendBackendModel.hasUnsavedScenes
+                providerResult = project.solution.frontendBackendModel.hasUnsavedState
                     .sync(Unit, RpcTimeouts(200L, 200L))
```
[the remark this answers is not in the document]
van800: I have tested it by generating 7000 assets with ScriptableObject and the call is so fast, that it is unrecognizable in the profiler.
When I tried to create 70000 assets, Unity got stuck processing them.

unity/EditorPlugin/AfterUnity56/Initialization.cs
```diff
@@ -71,7 +71,7 @@ public class SpawnManagerScriptableObject : ScriptableObject
         public Vector3[] spawnPoints;
     }
                  */
-                var hasDirtyUserAssets = Resources.FindObjectsOfTypeAll<Object>()
+                var hasDirtyUserAssets = Resources.FindObjectsOfTypeAll<ScriptableObject>()
```
citizenmatt, line 79: Does this check only apply to `ScriptableObject`? I guess other objects should belong to the scene. What about prefabs?

unity/EditorPlugin/AfterUnity56/Initialization.cs
```diff
@@ -77,14 +77,13 @@ public class SpawnManagerScriptableObject : ScriptableObject
...
-                        if (string.IsNullOrEmpty(assetPath))
+                        if (a.hideFlags < HideFlags.DontSaveInEditor) // this is faster than checking the File.Exists(Path.GetFullPath(assetPath))
                             return false;
-                        
+
                         if (!IsDirty(a))
                             return false;
 
-                        return File.Exists(Path.GetFullPath(assetPath));
+                        return true;
```
citizenmatt: Could simplify this to `return IsDirty(a);` but that's a really Minor nitpick

citizenmatt (approved):
LGTM

unity/EditorPlugin/AfterUnity56/Initialization.cs
van800: I made some experiments and it showed that prefabs are auto-saved.

van800:
merged https://buildserver.labs.intellij.net/buildConfiguration/DotNetDiv_Wave223_TriggerPullRequestMergeIntoNet223/201463110?buildTab=changes&showFiles=false&expandRevisionsSection=true

PR #2368, opened 2022-10-27
Mute InputSystem tests

rwx788:
We need to investigate what causes "Should not add/remove to packages without setting packagesUpdating" exception. Not to block other work, tests are temporaly muted.

See [RIDER-84142](https://youtrack.jetbrains.com/issue/RIDER-84142).


citizenmatt:
LGTM. I think I have a fix for the problem. Essentially, it's a false negative. The assert shouldn't be there, and everything is working correctly, apart from the message. I'll try and pull a fix together ASAP.

rwx788:
> LGTM. I think I have a fix for the problem. Essentially, it's a false negative. The assert shouldn't be there, and everything is working correctly, apart from the message. I'll try and pull a fix together ASAP.

Awesome! If you already have a fix, then we don't need this PR, so once you have it, feel free to close. Cheers!

PR #2369, opened 2022-10-27
Add check for struct based elements to UsageInspectionsSuppressor

jorgeamado:
Add usage suppression to structed based ISystem

`public readonly struct DebugVisualizationSystem : ISystem
{...}
`

https://youtrack.jetbrains.com/issue/RIDER-80062


jorgeamado:
Will be merged via TeamCity pipeline

PR #2370, opened 2022-10-28
Net223 rider-79869 IBaker Usage inscpection suppression

jorgeamado:
https://youtrack.jetbrains.com/issue/RIDER-84159/Class-derived-from-Baker-should-be-marked-as-used

```
using Unity.Entities;
using Unity.Rendering;

class CannonBallAuthoring : UnityEngine.MonoBehaviour
{
}

class CannonBallBaker : Baker<CannonBallAuthoring>
{
    public override void Bake(CannonBallAuthoring authoring)
    {
        // By default, components are zero-initialized.
        // So in this case, the Speed field in CannonBall will be float3.zero.
        AddComponent<CannonBall>();
        AddComponent<URPMaterialPropertyBaseColor>();
    }
}
```

### DIFF
--- a/resharper/resharper-unity/src/Unity/CSharp/Daemon/UsageChecking/UsageInspectionsSuppressor.cs
+++ b/resharper/resharper-unity/src/Unity/CSharp/Daemon/UsageChecking/UsageInspectionsSuppressor.cs
@@ -59,7 +59,10 @@ namespace JetBrains.ReSharper.Plugins.Unity.CSharp.Daemon.UsageChecking
                                      IsUxmlFactory(cls):
                     flags = ImplicitUseKindFlags.InstantiatedNoFixedConstructorSignature;
                     return true;
-
+                case IStruct @struct when unityApi.IsUnityType(@struct) ||
+                                     UnityApi.IsDotsSystemType(@struct) :
+                    flags = ImplicitUseKindFlags.InstantiatedNoFixedConstructorSignature;
+                    return true;
                 case ITypeElement typeElement when unityApi.IsSerializableTypeDeclaration(typeElement):
                     // TODO: We should only really mark it as in use if it's actually used somewhere
                     // That is, it should be used as a field in a Unity type, or another serializable type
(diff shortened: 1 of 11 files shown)


jorgeamado:
Please review after https://github.com/JetBrains/resharper-unity/pull/2369

resharper/resharper-unity/src/Unity/CSharp/Daemon/UsageChecking/UsageInspectionsSuppressor.cs
```diff
@@ -59,7 +59,10 @@ public bool SuppressUsageInspectionsOnElement(IDeclaredElement element, out Impl
                                      IsUxmlFactory(cls):
                     flags = ImplicitUseKindFlags.InstantiatedNoFixedConstructorSignature;
                     return true;
-
+                case IStruct @struct when unityApi.IsUnityType(@struct) ||
```
krasnotsvetov, line 62: 1. Why only struct? System also could be class
2. Why are you setting implicit use for any "IsUnityType"?

resharper/resharper-unity/src/Unity/UnityEditorIntegration/Api/UnityApi.cs
```diff
@@ -55,7 +55,9 @@ public UnityApi(UnityVersion unityVersion, UnityTypeCache unityTypeCache, UnityT
             type != null && myUnityTypeCache.IsUnityType(type);
 
         public static bool IsDotsSystemType([NotNullWhen(true)] ITypeElement? typeElement) =>
```
krasnotsvetov, line 57: IsDotsSystemType -> IsDotsType (baker is not system)

krasnotsvetov (changes requested):

resharper/resharper-unity/src/Unity/CSharp/Daemon/UsageChecking/UsageInspectionsSuppressor.cs
jorgeamado: Class is processed on the line 55

jorgeamado:
Will be merged via TeamCity pipeline

PR #2388, opened 2023-02-10
Refactor editor plugin

citizenmatt:
Don't panic! This isn't quite as scary as it looks. This PR refactors the editor plugin, hopefully without changing behaviour too much. It's mostly renaming and extracting, moving classes around and making some more things `internal`. It will probably be easier to review each commit, which should be small and self-contained, and to open the final layout and see how it looks rather than trying to review the full combined diff.

The main changes are:

* Add a document explaining what the different builds of the plugin mean
* Rename projects to be more descriptive, based on supported Unity versions. The assembly names are not changed.
* Removes the undistributed Unity 5.5 build of the plugin
* Split the `After56` folder into `SinceUnity.5.6`, `SinceUnity.2019.2` and `UntilUnity.2017.3` folders. These folders are included/excluded as folders, instead of excluding single files. Defines are used more frequently around a few files.
* Extract protocol initialisation from `PluginEntryPoint` into `UnityEditorProtocol`. This new class is also responsible for setting up the subscriptions for the version specific features, so there is now a single location to register all subscriptions to the model (handlers can be inline or in other classes).
* Remove the call to queue onto the `MainThreadDispatcher` in protocol callbacks. **This is the only real change in behaviour in the PR**. We know that all protocol callbacks are in response either to a Unity callback (which we know will be on the main thread) or from the protocol (which we also know is on the main thread). The asserts have been left in place, just to be sure.


unity/EditorPlugin/PluginEntryPoint.cs
```diff
@@ -1,85 +1,83 @@
...
-using JetBrains.Rider.Model.Unity;
-using JetBrains.Rider.Unity.Editor.Logger;
-using JetBrains.Rider.Unity.Editor.NonUnity;
-using JetBrains.Rider.Unity.Editor.Utils;
 using UnityEditor;
 using UnityEditor.Callbacks;
 using Application = UnityEngine.Application;
 using Debug = UnityEngine.Debug;
 
+// DO NOT CHANGE NAMESPACE OR CLASSNAME!
```
van800, line 15: Do you propose, that we change the calls in the package to point this new namespace?

unity/EditorPlugin/PluginSettings.cs
```diff
@@ -262,7 +261,7 @@ private static void RiderPreferencesItem()
       {
         //UnityEditorInternal.InternalEditorUtility.OpenFileAtLineExternal(PluginEntryPoint.LogPath, 0);
         // works much faster than the commented code, when Rider is already started
-        PluginEntryPoint.OpenAssetHandler.OnOpenedAsset(PluginEntryPoint.LogPath, 0, 0);
```
van800, line 265: this can't be changed unfortunately
used from the package https://github.cds.internal.unity3d.com/unity/com.unity.ide.rider/blob/80e35461c5bed207859e3bda4d1006d9cb79afbc/Packages/com.unity.ide.rider/Rider/Editor/EditorPluginInterop.cs#L71

van800:
Would you please check all the reflection calls from the `EditorPluginInterop` of the package?

unity/EditorPlugin/PluginEntryPoint.cs
citizenmatt: No, the namespace for `EntryPoint` hasn't changed, and I'm not going to change it. It'll stay as it was and the existing reflection call will work as before. But I'll extend this comment to state that it's called from the package

unity/EditorPlugin/PluginEntryPoint.cs
```diff
@@ -226,9 +220,6 @@ private static void InitializeProtocol(Lifetime lifetime, string protocolInstanc
 
     internal static void InitForPluginLoadedFromAssets()
     {
-      if (ourInitialized)
```
krasnotsvetov, line 229: I think that flag prevent from double initialization during same AppDomain session: if you have package installed and you still have old Editor Plugin in Assets folder

unity/EditorPlugin/PluginEntryPoint.cs
```diff
@@ -310,22 +317,22 @@ private static void SetupAssemblyReloadEvents()
...
               }
+
               PlayModeSavedState = newPlayModeState;
             }
           });
         }
       };
 
-      AppDomain.CurrentDomain.DomainUnload += (sender, args) =>
+      // TODO: Why are we doing this? Why reset play mode when the AppDomain is being unloaded?
```
krasnotsvetov: https://github.com/JetBrains/resharper-unity/issues/520

https://github.com/JetBrains/resharper-unity/pull/609

unity/EditorPlugin/PluginEntryPoint.cs
```diff
@@ -145,75 +145,74 @@ private static void Init(Lifetime lifetime)
...
+        }
+        else
+        {
+          ourLogger.Verbose("Protocol will be recreated on next domain load, plugin lifetime is not alive");
+        }
+      });
 
-        var result = ProtocolInstance.ToJson(protocols);
-        File.WriteAllText(protocolInstancePath, result);
+      ourLogger.Verbose("Writing Library/ProtocolInstance.json");
```
krasnotsvetov: let's move to first line of method or add new log: "start protocol initialization"

krasnotsvetov: Previously if somehting failed during scanning FS, we have last entry in log about protocol initialization

unity/EditorPlugin/PluginEntryPoint.cs
citizenmatt: If the plugin is loaded from `Assets`, Unity will invoke the class constructor, because it's `[InitializeOnLoad]`. This sets the flag. 

When the package is initialised, it looks to see if the plugin is already in memory, and if so, does nothing, so there won't be double initialisation.

I guess there could be a race condition here - if the package is loaded before the package, then the package would continue initialisation. But that would load the plugin from the IDE as a byte stream rather than a named assembly or file, which means the two `PluginEntryPoint` classes would be different `Type` instances, so the flag wouldn't help, right?

unity/EditorPlugin/PlayModeStateTracker.cs
```diff
@@ -0,0 +1,38 @@
...
+  {
+    public static readonly IViewableProperty<PlayModeState> Current = new ViewableProperty<PlayModeState>(GetPlayModeState());
+
+    public static void Initialise()
+    {
+#if UNITY_2017_3_OR_NEWER
+      EditorApplication.playModeStateChanged += _ => Current.Value = GetPlayModeState();
+#else
+      // playmodeStateChanged was marked obsolete in 2017.1, but it's still working in 2019.4
+      EditorApplication.playmodeStateChanged += () => Current.Value = GetPlayModeState();
```
krasnotsvetov, line 25: Interesting that we are not unsubscribing ever. It not your change, making comment just for history

unity/EditorPlugin/PluginEntryPoint.cs
```diff
@@ -273,58 +262,32 @@ private static void ResetDefaultFileExtensions()
...
 
     private static void SetupAssemblyReloadEvents(Lifetime lifetime)
     {
       // Unity supports recompile/reload settings natively for Unity 2018.2+
       if (UnityUtils.UnityVersion >= new Version(2018, 2))
         return;
 
-      // playmodeStateChanged was marked obsolete in 2017.1. Still working in 2018.3
-#pragma warning disable 618
-      EditorApplication.playmodeStateChanged += () =>
```
krasnotsvetov, line 302: Old subscription without unsubscribe

unity/EditorPlugin/PluginEntryPoint.cs
```diff
@@ -742,18 +710,7 @@ private static void AdviseUnityActions(BackendUnityModel model, Lifetime connect
         MainThreadDispatcher.Instance.Queue(EditorApplication.Step);
       });
 
-      var onPlaymodeStateChanged = new EditorApplication.CallbackFunction(() => syncPlayState());
-
-// left for compatibility with Unity <= 5.5
-#pragma warning disable 618
-      connectionLifetime.AddBracket(() => { EditorApplication.playmodeStateChanged += onPlaymodeStateChanged; },
```
krasnotsvetov, line 749: Old subscription WITH unsubscribe

unity/EditorPlugin/PluginEntryPoint.cs
citizenmatt: Thanks. I'll add a comment for context

unity/EditorPlugin/AfterUnity56/EntryPoint.cs
```diff
@@ -1,7 +1,4 @@
-﻿using JetBrains.Collections.Viewable;
-using JetBrains.Rider.Unity.Editor.FindUsages;
-using JetBrains.Rider.Unity.Editor.UnitTesting;
-using UnityEditor;
+﻿using UnityEditor;
 
 // DO NOT CHANGE NAMESPACE OR CLASSNAME!
 namespace JetBrains.Rider.Unity.Editor.AfterUnity56
```
krasnotsvetov: let's put disable R# inconsistent naming inspection

unity/EditorPlugin/PluginEntryPoint.cs
krasnotsvetov: I was reading commit one by one and see that you have refactored already it a bit. Yeah, now everything is correct (after reading the last commit). I suppose that previously in one of Rider Packages we were not checking that library was already loaded

citizenmatt: Cool, thanks. Marking as resolved

unity/EditorPlugin/PluginEntryPoint.cs
citizenmatt: Done

unity/EditorPlugin/PluginEntryPoint.cs
citizenmatt: Done

unity/EditorPlugin/PluginSettings.cs
citizenmatt: Done

citizenmatt:
Force pushed with changes to address comments. Added new commit before major changes (04d35c459a8bc4ad2734a15e59dc24f2939e9ea6) to add `[PublicAPI]` and a comment to all known type and type members that are accessed via reflection from the Unity editor package. Used this to trace through the rest of the commits to verify that all externally used types/type members are still available.

Kicking off a merge on TeamCity.

PR #2346, opened 2022-08-22
Disable API verification for Unity.Shaders.dll

ulex:
Unity.Shaders is a pluggable part of Unity plugin. The Unity plugin can
be used without installed ReSharper C++, but ReSharper's API verifier
doesn't know that and report an error during installation. To workaround
this issue disable API verification for Unity.Shaders explicitly.

Please note, that I targeted this PR to net222, but this change also required in net223.

PR #2451, opened 2024-07-01
Disable API verification for Unity.Shaders.dll

ulex:
Similar to https://github.com/JetBrains/resharper-unity/pull/2346, but it was added to the wrong section in csproj and therefore accidentally removed. 

Unity.Shaders is a pluggable part of Unity plugin. The Unity plugin can be used without installed ReSharper C++, but ReSharper's API verifier doesn't know that and report an error during installation. To workaround this issue disable API verification for Unity.Shaders explicitly.


citizenmatt (approved):
LGTM. Thanks!

citizenmatt:
Merged to `master` via TeamCity

